PR #28, opened 2018-02-09
[solution open for discussion] Fix error on PKCS8 keys

jgf:
Hi,

we noticed that trying to create a certificate object with a private key in PKCS8 format results in a 500 response "An application error occurred. Please contact your CredHub administrator".

On the server side an exception is thrown:
```
javax.validation.ValidationException: HV000028: Unexpected exception during isValid call.
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateSingleConstraint(ConstraintTree.java:451) ~[hibernate-validator-5.3.6.Final.jar:5.3.6.Final]
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateConstraints(ConstraintTree.java:127) ~[hibernate-validator-5.3.6.Final.jar:5.3.6.Final]
...
Caused by: java.lang.ClassCastException: org.bouncycastle.asn1.pkcs.PrivateKeyInfo cannot be cast to org.bouncycastle.openssl.PEMKeyPair
	at org.cloudfoundry.credhub.util.PrivateKeyReader.getPublicKey(PrivateKeyReader.java:42) ~[bin/:?]
```

This pull request fixes the error and returns a "The provided key format is not supported." response.

However I am unsure if we want to support PKCS8 key format instead of returning an error. Whats your opinion?

Best regards,
Jürgen

### DIFF
--- a/src/main/resources/messages.properties
+++ b/src/main/resources/messages.properties
@@ -27,6 +27,7 @@ error.invalid_duration=Invalid duration specified. The supported duration values
 error.invalid_json_key=The request includes an unrecognized parameter ''{0}''. Please update or remove this parameter and retry your request.
 error.invalid_key_length=The provided key length is not supported. Valid values include '2048', '3072' and '4096'.
 error.invalid_token_signature=The request token signature could not be verified. Please validate that your request token was issued by the UAA server authorized by CredHub.
+error.invalid_key_format=The provided key format is not supported.
 error.mismatched_certificate_and_private_key=The provided certificate does not match the private key.
 error.missing_certificate_credentials=At least one certificate attribute must be set. Please validate your input and retry your request.
 error.missing_certificate_parameters=At least one subject value, such as common name or organization, must be defined to generate the certificate. Please update and retry your request.
--- a/src/main/java/org/cloudfoundry/credhub/util/PrivateKeyReader.java
+++ b/src/main/java/org/cloudfoundry/credhub/util/PrivateKeyReader.java
@@ -13,16 +13,33 @@ import java.security.PublicKey;
 
 public class PrivateKeyReader {
 
-  public static PrivateKey getPrivateKey(String privateKeyPem) throws IOException {
+    public static class UnsupportedFormatException extends Exception {
+        private static final long serialVersionUID = -2669429797326574839L;
+        public UnsupportedFormatException(String msg) {
+            super(msg);
+        }
+    }
+
+  public static PrivateKey getPrivateKey(String privateKeyPem) throws IOException, UnsupportedFormatException {
     PEMParser pemParser = new PEMParser(new StringReader(privateKeyPem));
-    PEMKeyPair pemKeyPair = (PEMKeyPair) pemParser.readObject();
+    Object parsed = pemParser.readObject();
+    pemParser.close();
+    if (!(parsed instanceof PEMKeyPair)) {
+      throw new UnsupportedFormatException("format of private key is not supported.");
+    }
+    PEMKeyPair pemKeyPair = (PEMKeyPair) parsed;
     PrivateKeyInfo privateKeyInfo = pemKeyPair.getPrivateKeyInfo();
     return new JcaPEMKeyConverter().getPrivateKey(privateKeyInfo);
   }
 
-  public static PublicKey getPublicKey(String privateKeyPem) throws IOException {
+  public static PublicKey getPublicKey(String privateKeyPem) throws IOException, UnsupportedFormatException {
     PEMParser pemParser = new PEMParser(new StringReader(privateKeyPem));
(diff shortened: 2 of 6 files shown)


strehle:
Hi,
the cast exception shows, that it is PrivateKeyInfo
see here UAA code howto parse PEM encoded (encrypted) keys 
https://github.com/cloudfoundry/uaa/blob/master/server/src/main/java/org/cloudfoundry/identity/uaa/util/KeyWithCert.java#L93-L103

regards,
-markus

jgf:
Hi Markus,

thanks for the tip.

src/main/resources/messages.properties
```diff
@@ -27,6 +27,7 @@ error.invalid_duration=Invalid duration specified. The supported duration values
 error.invalid_json_key=The request includes an unrecognized parameter ''{0}''. Please update or remove this parameter and retry your request.
 error.invalid_key_length=The provided key length is not supported. Valid values include '2048', '3072' and '4096'.
 error.invalid_token_signature=The request token signature could not be verified. Please validate that your request token was issued by the UAA server authorized by CredHub.
+error.invalid_key_format=The provided key format is not supported.
```
voelzmo: It would probably be nice to say which formats are supported to help the user do "the right thing"

benmoss:
Hey @jgf , i haven't been able to reproduce the problem that this is intended to address. I tried running the `CertificateSetAndRegenerateTest.java` without the changes made to the implementation and it passed in spite of that, so I'm not sure it's providing the intended test coverage either.

Just manually, with the CredHub CLI, I tried running

```
openssl genpkey -algorithm RSA -pkeyopt rsa_keygen_bits:2048 -pkeyopt rsa_keygen_pubexp:65537 | openssl pkcs8 -topk8 -nocrypt -outform pem > rsa-2048-private-key.pem
credhub set -n foo -t certificate -p ./rsa-2048-private-key.pem
```

to create a PKCS8 key and put it into CredHub, and it appears to work fine for me. I tried the key you added to the test fixtures and that worked as well. Any idea why this might be?

voelzmo:
@benmoss people in your team should have certs to exactly reproduce this issue, see https://github.com/cloudfoundry-incubator/credhub/issues/26 for context.

jgf:
@benmoss thats weird. I can see the error every time when running `./gradlew --no-daemon -Dtest.single=CertificateSetAndRegenerateTest clean test`.

The credhub cli call probably did not fail in your case, because you did not specify a certificate and the relevant part of the validation is aborted (https://github.com/cloudfoundry-incubator/credhub/blob/master/src/main/java/org/cloudfoundry/credhub/validator/CertificateMatchesPrivateKeyValidator.java#L34).

Please try to checkout the `pkcs8_test` branch from [my fork](https://github.com/jgf/credhub/tree/pkcs8_test). You should be able to see an error running the command below
```
cat <<EOF >Print.java
class PrintKey {
  public static void main(String args[]) {
    System.out.println(org.cloudfoundry.credhub.util.TestConstants.TEST_PRIVATE_KEY_PKCS8);
  }
}

class PrintCert {
  public static void main(String args[]) {
    System.out.println(org.cloudfoundry.credhub.util.TestConstants.TEST_CERTIFICATE);
  }
}
EOF
javac Print.java src/test/java/org/cloudfoundry/credhub/util/TestConstants.java
java -cp src/test/java/:. PrintKey > key.pem
java -cp src/test/java/:. PrintCert > cert.pem
# Fails with: An application error occurred. Please contact your CredHub administrator.
credhub set -n test -t certificate -c cert.pem  -p key.pem
```

When you convert the key back to rsa format credhub should be able to handle it.
```
openssl rsa -in key.pem -text > rsa_key.pem
credhub set -n test -t certificate -c cert.pem  -p rsa_key.pem
```

So thats how I can reproduce the error. However I still have no clue, why it did not work initially in your case.

benmoss:
Ok, I retried with a `gradle clean` and it did indeed fail! Must have been that I changed the files with Git and IntelliJ hadn't noticed the changes. Sorry about that. Thanks for the other repro too, I'll take a further look today.

benmoss:
And it looks like the reason it didn't fail when I tried my pkcs8 key was because this code is only executed when you set a certificate *and* a pkcs8 key. It is a validation we put in to ensure that the key  matches the certificate.

benmoss:
I think because of a8cd22e994ec3a5fceedf3ef3641985c6522266f / https://www.pivotaltracker.com/story/show/152650210 we are in a position where we need to be able to derive the public key from the private key, which from what I understand is possible with RSA private keys but not universally true of all private keys. 

Not a crypto expert, so this is a bit over my head, but it might have been short-sighted to implement that feature because it may limit our support for different key types. I think in the short-term erroring better is the best we can do, but longer term we should see if we can derive public keys for certificate verification from *all* valid private keys, or else possibly conditionally drop the certificate validation we added.

PR #29, opened 2018-02-14
Add secretKeyMode to password generation

tn5993:
Signed-off-by: Amanda Alvarez <amanda.alvarez@emc.com>

This is our first attempt at this, so we would appreciate feedback on how to improve our contribution.
We would like Credhub to generate a secret key in Base64 with the following format:
```
---
variables:
- name: /some-secret
  type: password
  options:
    length: 40 # this length will be byte length when secret_key_mode is on
    secret_key_mode: true
```
We modified some of the tests, and all tests passed.

### DIFF
--- a/src/main/java/org/cloudfoundry/credhub/generator/PasswordCredentialGenerator.java
+++ b/src/main/java/org/cloudfoundry/credhub/generator/PasswordCredentialGenerator.java
@@ -1,10 +1,15 @@
 package org.cloudfoundry.credhub.generator;
 
+import org.apache.commons.codec.binary.Base64;
 import org.cloudfoundry.credhub.credential.StringCredentialValue;
 import org.cloudfoundry.credhub.request.GenerationParameters;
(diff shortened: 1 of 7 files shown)


crawsible:
Thanks for the PR @tn5993 and Amanda! We will review this in the coming days. We will likely ask you to change around the interface a bit; "secret_key_mode" is a little too AWS-specific of a name, and we'd prefer to be generic about the values we generate. But no action needed on your part yet, we'll look over this in due time and give you feedback then. Thanks!

crawsible:
Thanks again for the PR. We checked this out and we still feel like the close ties to S3 naming conventions for something that is actually quite generic is the wrong approach. Can you revise your PR so that the api accepts `"base64": true` instead of `"secret_key_mode"`, and update the code names accordingly?

mdelillo:
Closing due to lack of response. Feel free to reopen with the requested changes.

PR #42, opened 2018-08-08
Support all valid JSON

LinuxBozo:
This PR allows CredHub to support any valid JSON, not just map. Requires a change to the CLI as well to read the values once set.

One part of the fix for cloudfoundry-incubator/credhub-cli#30


martys34:
Your changes look good so we merged them. They should be out in the next CredHub release.

PR #61, opened 2019-11-13
Fixes for docker-compose setup

cholick:
* In security bootstrapping, actors is an array.
* uaa requires a trailing / to correctly work (credhub CLI handles it ok with the slash, but the other library I was using didn't add the slash coming back from the /info endpoint and wasn't able to successfully login)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,8 +28,8 @@ services:
     ports:
     - 9000:9000
     environment:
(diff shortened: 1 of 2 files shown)


cholick:
(Please squash for the merge)

ankeesler:
Thank you for this @cholick !

Issue #134, opened 2020-06-08
Importing a self-signed certificate not possible


dueckminor:
What version of the credhub server you are using?
2.6.0

What version of the credhub cli you are using?
2.6.2

If you were attempting to accomplish a task, what was it you were attempting to do?
```
> credhub generate -n  /selfsigned/cert  --type=certificate --common-name="selfsigned"  --self-sign
> credhub export --path=/selfsigned | credhub import --file=/dev/stdin
Credential '/selfsigned/cert' at index 0 could not be set: The provided CA value is not a valid X509 certificate authority.

Import complete.
Successfully set: 0
Failed to set: 1
 - Credential '/selfsigned/cert' at index 0 could not be set: The provided CA value is not a valid X509 certificate authority.
One or more credentials failed to import.
```
What did you expect to happen?
I would expect, that I can import things I have created using the CredHub-CLI and that they look exactly the same afterwards (ignoring the changed id)

What was the actual behavior?
The self signed certificate can not be imported

Please confirm where necessary:
* [ ] I have included a log output
* [ ] My log includes an error message
* [X] I have included steps for reproduction

If you are a PCF customer with an Operation Manager (PCF Ops Manager) please direct your questions to support (https://support.pivotal.io/)

bruce-ricard:
Note to whomever will fix this issue: when you think you are done, please add a test in our [import-export integration tests](https://github.com/cloudfoundry-incubator/credhub-acceptance-tests/blob/7a643ca064f9224b75ae25a341245a84c6e595ba/integration_test/import_export_test.go#L98-L101) to deal with self-signed leaves.

HappyTobi:
Hi all,

The CredHub and CredHub acceptance test was done

CredHub PR: #150 
CredHub acceptance test PR: [#7](https://github.com/cloudfoundry-incubator/credhub-acceptance-tests/pull/7)

Kind regards

jknostman3:
@HappyTobi & @damzog I am happy to report that we have made a new release of CredHub (2.9.0) that includes this fix. As a result, I will close out this issue.

PR #139, opened 2020-06-16
FIX Generate self signed certificate without the ca flag #134

HappyTobi:
Hi all,

I just fixed the issue mentioned in #134 

So from my perspective it was not correct to export a Certificate as a CA when you didn't pass the `is_ca` flag or a valid (known) ca that should be used to sign the certificate.

So the fix was just looking into the arguments and only return a certificate in CA field if there a `known` ca or the `is_ca` flag was passed.

Hope you like that fix.

Kind regards

### DIFF
--- a/components/credentials/src/test/java/org/cloudfoundry/credhub/generators/CertificateGeneratorTest.java
+++ b/components/credentials/src/test/java/org/cloudfoundry/credhub/generators/CertificateGeneratorTest.java
@@ -264,7 +264,7 @@ public class CertificateGeneratorTest {
       equalTo(CertificateFormatter.pemOf(rootCaKeyPair.getPrivate())));
     assertThat(certificateCredential.getCertificate(),
       equalTo(CertificateFormatter.pemOf(certificate)));
-    assertThat(certificateCredential.getCa(), equalTo(CertificateFormatter.pemOf(certificate)));
+    assertThat(certificateCredential.getCa(), equalTo(""));
     verify(signedCertificateGenerator, times(1)).getSelfSigned(rootCaKeyPair, inputParameters);
   }
 
--- a/backends/credhub/src/test/java/org/cloudfoundry/credhub/integration/CertificateGenerateTest.java
+++ b/backends/credhub/src/test/java/org/cloudfoundry/credhub/integration/CertificateGenerateTest.java
@@ -1,6 +1,7 @@
 package org.cloudfoundry.credhub.integration;
 
 import java.io.ByteArrayInputStream;
+import java.io.StringReader;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.time.temporal.ChronoUnit;
@@ -8,6 +9,9 @@ import java.util.Calendar;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.bouncycastle.asn1.x509.*;
+import org.bouncycastle.cert.X509CertificateHolder;
+import org.bouncycastle.openssl.PEMParser;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.test.annotation.DirtiesContext;
@@ -22,9 +26,6 @@ import org.springframework.web.context.WebApplicationContext;
 
 import com.google.common.collect.ImmutableMap;
 import com.jayway.jsonpath.JsonPath;
-import org.bouncycastle.asn1.x509.AuthorityKeyIdentifier;
-import org.bouncycastle.asn1.x509.Extension;
-import org.bouncycastle.asn1.x509.SubjectKeyIdentifier;
 import org.bouncycastle.cert.jcajce.JcaX509ExtensionUtils;
 import org.cloudfoundry.credhub.CredhubTestApp;
 import org.cloudfoundry.credhub.ErrorMessages;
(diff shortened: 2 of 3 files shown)


bruce-ricard:
Hello @HappyTobi .

Thank you for your PR!

You wrote:
> it was not correct to export a Certificate as a CA when you didn't pass the is_ca flag or a valid (known) ca that should be used to sign the certificate.

First of all, I believe that instead of
> the is_ca flag

you must have meant
> the --self-signed flag

didn't you?


Now, more importantly, I don't believe that I agree with you on that. A "self-signed certificate" is NOT unsigned, it is indeed self-signed, meaning signed by itself. 
If you try to curl an HTTPS endpoint of a webserver using a self-signed certificate, you'll first get an SSL verification error. If you add the `--cacert server.crt` option to your curl, where `server.crt` is your self-signed certificate (which also happens to be its own CA certificate), then the curl passes.

IMHO it makes a lot of sense to add the self-signed certificate in the `ca` field. If anything else, it's a lot easier for automation: if an SSL client wants to contact an SSL server using certificate `C`, just go get `C.ca` from Credhub, and add it to your Client's trusted CAs. Even-though the certificate says `CA: false` when you parse it, it appears that clients are happy to use it as a CA to validate server certificates. 

The Credhub server appears to enforce during an `import` that all CA certificates are `CA: true`, and maybe that's what we should rather fix.

Because of this, I don't believe we'd want to accept this PR. But I am always very happy to keep the conversation going, and hear from you if you have any thoughts or disagreements about what I wrote above.

Thank you again for your interest and your time.

components/credentials/src/test/java/org/cloudfoundry/credhub/generators/CertificateGeneratorTest.java
```diff
@@ -264,7 +264,7 @@ public void whenSelfSignIsTrue_itGeneratesAValidSelfSignedCertificate() throws E
       equalTo(CertificateFormatter.pemOf(rootCaKeyPair.getPrivate())));
     assertThat(certificateCredential.getCertificate(),
       equalTo(CertificateFormatter.pemOf(certificate)));
-    assertThat(certificateCredential.getCa(), equalTo(CertificateFormatter.pemOf(certificate)));
+    assertThat(certificateCredential.getCa(), equalTo(""));
```
bruce-ricard, line 267: I don't believe this change would be positive overall for Credhub users. It would also be a breaking one.

HappyTobi:
Hi @bruce-ricard 

Thanks for your Answer.
 
> Hello @HappyTobi .
> 
> Thank you for your PR!
> 
> You wrote:
> 
> > it was not correct to export a Certificate as a CA when you didn't pass the is_ca flag or a valid (known) ca that should be used to sign the certificate.
> 
> First of all, I believe that instead of
> 
> > the is_ca flag
> 
> you must have meant
> 
> > the --self-signed flag
> 
> didn't you?

No, when you are just using the `--self-signed` flag, your cert will be signed by the cert itself.
(But the cert has no `CA:true` flag as property)

With the usage of
`--self-signed` and `is_ca` the cert will be also signed with the cert it self, but it will also get the `CA:true` flag.

> 
> Now, more importantly, I don't believe that I agree with you on that. A "self-signed certificate" is NOT unsigned, it is indeed self-signed, meaning signed by itself.
> If you try to curl an HTTPS endpoint of a webserver using a self-signed certificate, you'll first get an SSL verification error. If you add the `--cacert server.crt` option to your curl, where `server.crt` is your self-signed certificate (which also happens to be its own CA certificate), then the curl passes.

Yes that's correct (sorry if  my comment was misunderstanding)

> 
> IMHO it makes a lot of sense to add the self-signed certificate in the `ca` field. If anything else, it's a lot easier for automation: if an SSL client wants to contact an SSL server using certificate `C`, just go get `C.ca` from Credhub, and add it to your Client's trusted CAs. Even-though the certificate says `CA: false` when you parse it, it appears that clients are happy to use it as a CA to validate server certificates.

Thats true! It's not needed for validating certificates or a certificate chain, but if you read the **ietf** documentation the `ca` certificate should contain the `ca:true` constraint to be **conforming** CA certificate.  (That's what I understand when I read the documentation)
See:
https://tools.ietf.org/html/rfc5280#section-4.2.1.2

> 
> The Credhub server appears to enforce during an `import` that all CA certificates are `CA: true`, and maybe that's what we should rather fix.
>

Think that's a point you should think of. 
If you say that all conforming CA Certificates have the `CA:true` attribute then the check of today is correct and we should change the exports else you have to think of.


> Because of this, I don't believe we'd want to accept this PR. But I am always very happy to keep the conversation going, and hear from you if you have any thoughts or disagreements about what I wrote above.
> 
> Thank you again for your interest and your time.
You're welcome.

I'm also happy to hear something from you about my comments.

Kind regards

bruce-ricard:
@HappyTobi , thank you for your quick reply.

Thanks for the thoughts and the link to the RFC document (it's interesting but I'll admit I didn't read all of it :))

After looking at the spec you shared, it looks like you are right in saying that a CA needs to be a "conforming" CA, and that in the case of a self-signed certificate, what signed it (i.e. itself) is by definition not a "conforming CA", since it's a leaf. 
We also looked at how `openssl` generates self-signed certificates, and we noticed that they don't have a CA field in them, whereas Credhub-generated self-signed certificates do have a CA field containing essentially themselves.

Credhub doesn't attempt to be a low level implementation of the spec, but tries to be at a higher level of abstraction, at which every certificate has a filled CA field. If we make a change that nulls out some CA values, we break this contract. That is the reason why we'd rather not null out any CA value.

A rationalization for this decision could be:
The idea behind Credhub is to provide an abstraction and a tool on top of the quite complex PKI and low level tools (like `openssl`). The `ca` field in Credhub maybe could be rather seen as a `signer` field which represents an abstraction that can represent either a real CA or the certificate itself, when it's self-signed.

My team spent some time brainstorming this issue and came out with this proposed solution to the problem:
> At the `import` level, when attempting to validate a certificate, if the `ca` field is the same as the `certificate` field, do *not* attempt to validate that it is a conforming CA certificate.

The reason why we liked this solution is because:
1. It is not a breaking change
1. The change should be quite easy 
1. The abstraction of the `ca` field always being whatever signed the certificate is kept

We are currently quite short staffed and do not have the bandwidth to work on this. We would greatly appreciate it if you could submit another PR implementing this solution.

Once again, feel free to brainstorm that solution with us if you believe it is flawed in any way. 

Thank you,
The Credhub team @cloudfoundry-incubator/cf-credhub

jknostman3:
We are closing this PR in light of the new PR based on the comments above. Check here for more details: https://github.com/cloudfoundry-incubator/credhub/pull/150

PR #165, opened 2020-11-17
Define abbreviations at their first use

thausler786:
[174322077](https://www.pivotaltracker.com/story/show/174322077)


peterhaochen47:
Looks good, thanks for the clarification!

PR #183, opened 2021-04-01
Upgrade flyway

jpalermo:
This allows support of postgres 13 which has been added to the bosh director

The issue with flyway v4 having a different default schema migrations table name
than flyway v5+ is resolved with a check that runs before migrating, and if the
old table is found, it is simply renamed to the new table.

We were able to run the application in dev mode against postgres just fine, and saw it migrate from the old table name to the new. But when we run tests, there seems to be some errors that show up because of possible changes in flyway that are not supported in the current version of spring boot.

We made an attempt to upgrade spring boot to resolve these errors, but we quickly got out of our element and rolled that work back.

Since the app seems to run, it may be possible to resolve the problem in how the tests are run instead of with a spring boot upgrade (although it does seem like it's time for a spring boot upgrade anyway).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ buildscript {
         bcpkixFipsVersion = '1.0.2'
         bcFipsVersion = '1.0.1'
         commonsCodecVersion = '1.14' // remove this after deleting (now deprecated) spring-security-oauth2
-        flywayVersion = '5.2.4' // pinning to v5 because upgrading to v6+ directly breaks the upgrade of (v4->v5->v6) unless we can work around [this breaking change](https://stackoverflow.com/questions/49063385/flyway-5-0-7-warning-about-using-schema-version-table).
+        flywayVersion = '7.7.1'
         guavaVersion = '29.0-jre'
         h2Version = '1.4.199'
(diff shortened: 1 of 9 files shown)


build.gradle
```diff
@@ -9,7 +9,7 @@ buildscript {
         bcpkixFipsVersion = '1.0.2'
         bcFipsVersion = '1.0.1'
         commonsCodecVersion = '1.14' // remove this after deleting (now deprecated) spring-security-oauth2
-        flywayVersion = '5.2.4' // pinning to v5 because upgrading to v6+ directly breaks the upgrade of (v4->v5->v6) unless we can work around [this breaking change](https://stackoverflow.com/questions/49063385/flyway-5-0-7-warning-about-using-schema-version-table).
```
bruce-ricard, line 12: Did you look at this @jpalermo ? ~6 months ago we tried upgrading to flyway 6 and we had issues.

rkoster:
@jpalermo It seems like there are some merge conflicts. Is this PR still needed? If so could you rebase your changes?

jpalermo:
Looks like this was already merged (but rebased so github missed it). Thanks

PR #201, opened 2021-07-23
Add logic to generate certificates using minimum durations

pvaramballypivot:
- Receive 2 new fields in the server configuration file: ca_minimum_duration and leaf_minimum_duration
- Update cert generation/regeneration logic to use the values in these fields such that the cert gets the larger duration between the requested and the new fields
- Update converge mode logic to no longer consider the duration field
- Return a duration_overridden flag in the generate/regenerate API response when the minimum duration is used

For more information, refer this issue: https://github.com/pivotal/credhub-release/issues/60

cc @pabloarodas @ystros

### DIFF
--- a/applications/credhub-api/src/test/resources/application-minimum-duration.yml
+++ b/applications/credhub-api/src/test/resources/application-minimum-duration.yml
@@ -1,0 +1,3 @@
+certificates:
+  ca_minimum_duration: 1825
+  leaf_minimum_duration: 1460
--- a/applications/credhub-api/src/test/resources/application-minimum-duration.yml
+++ b/applications/credhub-api/src/test/resources/application-minimum-duration.yml
@@ -1,3 +1,4 @@
 certificates:
-  ca_minimum_duration: 1825
-  leaf_minimum_duration: 1460
+  ca_minimum_duration_in_days: 1825
+  leaf_minimum_duration_in_days: 1460
+
--- a/backends/credhub/src/docs/asciidoc/snippets/introduction.adoc
+++ b/backends/credhub/src/docs/asciidoc/snippets/introduction.adoc
@@ -36,3 +36,5 @@ Credential responses include a unique identifier in the key 'id'. This ID is a u
 As of 2.0.0, set requests always overwrite the credential that already exists.
 
 As of 2.0.1, generate requests can be set to overwrite, no-overwrite, or converge for the mode parameter. The default mode for generate is converge as of 2.0.0. Converge will only overwrite if the generate request parameters do not match the existing credential.
+
+As of 2.10.0, when the generate requests are set to converge for the mode parameter, converge will not overwrite certificates if duration is the only parameter that does not match the existing certificate credentials.
--- a/backends/credhub/src/test/java/org/cloudfoundry/credhub/handlers/DefaultRegenerateHandlerTest.java
+++ b/backends/credhub/src/test/java/org/cloudfoundry/credhub/handlers/DefaultRegenerateHandlerTest.java
@@ -185,7 +185,7 @@ public class DefaultRegenerateHandlerTest {
       .thenReturn(savedCredentialVersion);
 
     CredentialView actualCredentialView = subjectWithAclsEnabled.handleRegenerate(CREDENTIAL_NAME, null);
-    CredentialView expectedCredentialView = CredentialView.fromEntity(savedCredentialVersion);
+    CredentialView expectedCredentialView = CredentialView.fromEntity(savedCredentialVersion, false, true);
 
     verify(permissionCheckingService, times(1)).hasPermission(USER, CREDENTIAL_NAME, PermissionOperation.WRITE);
     assertThat(actualCredentialView).isEqualTo(expectedCredentialView);
@@ -291,7 +291,7 @@ public class DefaultRegenerateHandlerTest {
       .thenReturn(savedCredentialVersion);
 
     CredentialView actualCredentialView = subjectWithAclsDisabled.handleRegenerate(CREDENTIAL_NAME, null);
-    CredentialView expectedCredentialView = CredentialView.fromEntity(savedCredentialVersion);
+    CredentialView expectedCredentialView = CredentialView.fromEntity(savedCredentialVersion, false, true);
 
     verify(permissionCheckingService, times(0)).hasPermission(USER, CREDENTIAL_NAME, PermissionOperation.WRITE);
     assertThat(actualCredentialView).isEqualTo(expectedCredentialView);
--- a/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
+++ b/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
@@ -56,12 +56,10 @@ public class CertificateGenerator implements CredentialGenerator<CertificateCred
     }
 
     final int originalDuration = params.getDuration();
-    Boolean durationOverridden = null;
 
-    if (isMinimumDurationEnabled()) {
-      params.setDuration(getCertificateDuration(params));
-      durationOverridden = (originalDuration != params.getDuration());
-    }
+    params.setDuration(getCertificateDuration(params));
+
+    boolean durationOverridden = (originalDuration != params.getDuration());
 
     if (params.isSelfSigned()) {
       try {
@@ -134,9 +132,5 @@ public class CertificateGenerator implements CredentialGenerator<CertificateCred
       return Math.max(params.getDuration(), caMinimumDurationInDays);
     }
     return Math.max(params.getDuration(), leafMinimumDurationInDays);
-  }
-
-  private boolean isMinimumDurationEnabled() {
-    return (caMinimumDurationInDays !=0 || leafMinimumDurationInDays != 0);
   }
 }
--- a/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
+++ b/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
@@ -4,8 +4,6 @@ import java.security.KeyPair;
 import java.security.cert.X509Certificate;
 import java.time.Instant;
 
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
@@ -50,8 +48,6 @@ public class CertificateGenerator implements CredentialGenerator<CertificateCred
     final CertificateGenerationParameters params = (CertificateGenerationParameters) p;
     final KeyPair keyPair;
     final String privatePem;
-    final int originalDuration = params.getDuration();
-    Boolean durationOverridden = null;
     try {
       keyPair = keyGenerator.generateKeyPair(params.getKeyLength());
       privatePem = CertificateFormatter.pemOf(keyPair.getPrivate());
@@ -59,11 +55,10 @@ public class CertificateGenerator implements CredentialGenerator<CertificateCred
       throw new RuntimeException(e);
     }
 
-    if (params.isCa() && caMinimumDuration != null) {
-      params.setDuration(Math.max(params.getDuration(), caMinimumDuration));
-    } else if (!params.isCa() && leafMinimumDuration != null) {
-      params.setDuration(Math.max(params.getDuration(), leafMinimumDuration));
-    }
+    final int originalDuration = params.getDuration();
+    Boolean durationOverridden = null;
+
+    params.setDuration(getCertificateDuration(params));
 
     if (originalDuration != params.getDuration()) {
       durationOverridden = true;
@@ -133,5 +128,14 @@ public class CertificateGenerator implements CredentialGenerator<CertificateCred
     }
 
     return transitionalVersionCreatedAt.isAfter(latestNonTransitionalVersionCreatedAt);
+  }
+
+  private int getCertificateDuration(final CertificateGenerationParameters params) {
+    if (params.isCa() && caMinimumDuration != null) {
+      return Math.max(params.getDuration(), caMinimumDuration);
+    } else if (!params.isCa() && leafMinimumDuration != null) {
+      return Math.max(params.getDuration(), leafMinimumDuration);
+    }
+    return params.getDuration();
   }
 }
--- a/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
+++ b/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
@@ -58,10 +58,9 @@ public class CertificateGenerator implements CredentialGenerator<CertificateCred
     final int originalDuration = params.getDuration();
     Boolean durationOverridden = null;
 
-    params.setDuration(getCertificateDuration(params));
-
-    if (originalDuration != params.getDuration()) {
-      durationOverridden = true;
+    if (isMinimumDurationEnabled()) {
+      params.setDuration(getCertificateDuration(params));
+      durationOverridden = (originalDuration != params.getDuration());
     }
 
     if (params.isSelfSigned()) {
@@ -135,5 +134,9 @@ public class CertificateGenerator implements CredentialGenerator<CertificateCred
       return Math.max(params.getDuration(), caMinimumDurationInDays);
     }
     return Math.max(params.getDuration(), leafMinimumDurationInDays);
+  }
+
+  private boolean isMinimumDurationEnabled() {
+    return (caMinimumDurationInDays !=0 || leafMinimumDurationInDays != 0);
   }
 }
--- a/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
+++ b/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
@@ -64,7 +64,7 @@ public class CertificateGenerator implements CredentialGenerator<CertificateCred
     if (params.isSelfSigned()) {
       try {
         final String cert = CertificateFormatter.pemOf(signedCertificateGenerator.getSelfSigned(keyPair, params));
-        return new CertificateCredentialValue(cert, cert, privatePem, null, null, params.isCa(), params.isSelfSigned(), true, false, durationOverridden);
+        return new CertificateCredentialValue(cert, cert, privatePem, null, null, params.isCa(), params.isSelfSigned(), true, false, durationOverridden, params.getDuration());
       } catch (final Exception e) {
         throw new RuntimeException(e);
       }
@@ -102,7 +102,7 @@ public class CertificateGenerator implements CredentialGenerator<CertificateCred
           certificateReader.getCertificate(),
           PrivateKeyReader.getPrivateKey(signingCaPrivateKey)
         );
-        return new CertificateCredentialValue(signingCaCertificate, CertificateFormatter.pemOf(cert), privatePem, caName, trustedCaCertificate, params.isCa(), params.isSelfSigned(), true, false, durationOverridden);
+        return new CertificateCredentialValue(signingCaCertificate, CertificateFormatter.pemOf(cert), privatePem, caName, trustedCaCertificate, params.isCa(), params.isSelfSigned(), true, false, durationOverridden, params.getDuration());
       } catch (final Exception e) {
         throw new RuntimeException(e);
       }
--- a/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
+++ b/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
@@ -4,6 +4,8 @@ import java.security.KeyPair;
 import java.security.cert.X509Certificate;
 import java.time.Instant;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
@@ -20,6 +22,8 @@ import org.cloudfoundry.credhub.utils.PrivateKeyReader;
 
 @Component
 public class CertificateGenerator implements CredentialGenerator<CertificateCredentialValue> {
+
+  private static final Logger LOGGER = LogManager.getLogger(CertificateGenerator.class.getName());
 
   private final RsaKeyPairGenerator keyGenerator;
   private final SignedCertificateGenerator signedCertificateGenerator;
@@ -48,6 +52,7 @@ public class CertificateGenerator implements CredentialGenerator<CertificateCred
     final CertificateGenerationParameters params = (CertificateGenerationParameters) p;
     final KeyPair keyPair;
     final String privatePem;
+    final int originalDuration = params.getDuration();
     Boolean durationOverridden = null;
     try {
       keyPair = keyGenerator.generateKeyPair(params.getKeyLength());
@@ -58,10 +63,13 @@ public class CertificateGenerator implements CredentialGenerator<CertificateCred
 
     if (params.isCa() && caMinimumDuration != null) {
       params.setDuration(Math.max(params.getDuration(), caMinimumDuration));
-      durationOverridden = true;
     } else if (!params.isCa() && leafMinimumDuration != null) {
       params.setDuration(Math.max(params.getDuration(), leafMinimumDuration));
+    }
+
+    if (originalDuration != params.getDuration()) {
       durationOverridden = true;
+      LOGGER.info("Certificate duration overridden, original duration: {} - duration used: {}", originalDuration, params.getDuration());
     }
 
     if (params.isSelfSigned()) {
--- a/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
+++ b/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
@@ -48,6 +48,7 @@ public class CertificateGenerator implements CredentialGenerator<CertificateCred
     final CertificateGenerationParameters params = (CertificateGenerationParameters) p;
     final KeyPair keyPair;
     final String privatePem;
+    Boolean durationOverridden = null;
     try {
       keyPair = keyGenerator.generateKeyPair(params.getKeyLength());
       privatePem = CertificateFormatter.pemOf(keyPair.getPrivate());
@@ -57,14 +58,16 @@ public class CertificateGenerator implements CredentialGenerator<CertificateCred
 
     if (params.isCa() && caMinimumDuration != null) {
       params.setDuration(Math.max(params.getDuration(), caMinimumDuration));
+      durationOverridden = true;
     } else if (!params.isCa() && leafMinimumDuration != null) {
       params.setDuration(Math.max(params.getDuration(), leafMinimumDuration));
+      durationOverridden = true;
     }
 
     if (params.isSelfSigned()) {
       try {
         final String cert = CertificateFormatter.pemOf(signedCertificateGenerator.getSelfSigned(keyPair, params));
-        return new CertificateCredentialValue(cert, cert, privatePem, null, params.isCa(), params.isSelfSigned(), true, false);
+        return new CertificateCredentialValue(cert, cert, privatePem, null, null, params.isCa(), params.isSelfSigned(), true, false, durationOverridden);
       } catch (final Exception e) {
         throw new RuntimeException(e);
       }
@@ -102,7 +105,7 @@ public class CertificateGenerator implements CredentialGenerator<CertificateCred
           certificateReader.getCertificate(),
           PrivateKeyReader.getPrivateKey(signingCaPrivateKey)
         );
-        return new CertificateCredentialValue(signingCaCertificate, CertificateFormatter.pemOf(cert), privatePem, caName, trustedCaCertificate, params.isCa(), params.isSelfSigned(), true, false);
+        return new CertificateCredentialValue(signingCaCertificate, CertificateFormatter.pemOf(cert), privatePem, caName, trustedCaCertificate, params.isCa(), params.isSelfSigned(), true, false, durationOverridden);
       } catch (final Exception e) {
         throw new RuntimeException(e);
       }
--- a/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
+++ b/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
@@ -5,6 +5,7 @@ import java.security.cert.X509Certificate;
 import java.time.Instant;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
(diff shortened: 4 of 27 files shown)


backends/credhub/src/docs/asciidoc/snippets/introduction.adoc
```diff
@@ -36,3 +36,5 @@ Credential responses include a unique identifier in the key 'id'. This ID is a u
 As of 2.0.0, set requests always overwrite the credential that already exists.
 
 As of 2.0.1, generate requests can be set to overwrite, no-overwrite, or converge for the mode parameter. The default mode for generate is converge as of 2.0.0. Converge will only overwrite if the generate request parameters do not match the existing credential.
+
+As of 2.10.0, when the generate requests are set to converge for the mode parameter, converge will not overwrite certificates if duration is the only parameter that does not match the existing certificate credentials.
```
pvaramballypivot, line 40: We took a guess on the version the new feature will be in. This might need to be updated.

peterhaochen47:
Updating media type (496e70cc2e59f5a06b9601e35c5dc557fadfab90)

ystros:
> Updating media type ([496e70c](https://github.com/cloudfoundry-incubator/credhub/commit/496e70cc2e59f5a06b9601e35c5dc557fadfab90))

We can drop that commit if you think it's too unrelated (which is why we made it as a separate commit). We noticed when running the linter that it was complaining loudly about that due to Spring version, so fixed it as we were addressing other things. The linter warnings were present in the original code and was not caused by changes we made.

backends/credhub/src/docs/asciidoc/snippets/introduction.adoc
bruce-ricard: Two things:

1. This looks like a breaking change to me. If we ship it as is, I would say we need to bump major.
1. We are trying to stay on the "one version" idea, so that we don't have multiple versions to maintain and patch. I think that we would like this feature to be configurable through a feature flag so that we don't need to bump major and be forced to maintain multiple versions.

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
```diff
@@ -23,17 +24,23 @@
   private final RsaKeyPairGenerator keyGenerator;
   private final SignedCertificateGenerator signedCertificateGenerator;
   private final CertificateAuthorityService certificateAuthorityService;
+  private final Integer caMinimumDuration;
+  private final Integer leafMinimumDuration;
```
bruce-ricard, line 28: I would like to see these variables to be of type [Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html). 

In the configuration, we don't have an option and have to use Strings and Integer, but once the middleware is passed, I think that it's better to use types that represent better what the object is. 
It also solves the problem of the variable name not expressing which unit the integer is in.

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
```diff
@@ -23,17 +24,23 @@
...
-    final CertificateAuthorityService certificateAuthorityService) {
+    final CertificateAuthorityService certificateAuthorityService,
+    @Value("${certificates.ca_minimum_duration:#{null}}") final Integer caMinimumDuration,
+    @Value("${certificates.leaf_minimum_duration:#{null}}") final Integer leafMinimumDuration) {
     super();
     this.keyGenerator = keyGenerator;
     this.signedCertificateGenerator = signedCertificateGenerator;
     this.certificateAuthorityService = certificateAuthorityService;
+    this.caMinimumDuration = caMinimumDuration;
+    this.leafMinimumDuration = leafMinimumDuration;
```
bruce-ricard, line 43: And on these lines, the values would be converted form the configuration Integer to the backend's Duration.

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
```diff
@@ -48,10 +55,19 @@ public CertificateCredentialValue generateCredential(final GenerationParameters
...
+
+    if (originalDuration != params.getDuration()) {
+      durationOverridden = true;
+    }
+
     if (params.isSelfSigned()) {
       try {
         final String cert = CertificateFormatter.pemOf(signedCertificateGenerator.getSelfSigned(keyPair, params));
-        return new CertificateCredentialValue(cert, cert, privatePem, null, params.isCa(), params.isSelfSigned(), true, false);
+        return new CertificateCredentialValue(cert, cert, privatePem, null, null, params.isCa(), params.isSelfSigned(), true, false, durationOverridden);
```
bruce-ricard, line 70: It appears that you create a variable of type `Boolean`, but you never use the value `false`.

If I read this correctly, you initialize it to `null`, optionally set it to `true` on line 64, but no code ever sets it to `false`.

Am I reading this correctly? 
If yes, then maybe `Boolean` isn't the type you want to use.

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/domain/CertificateGenerationParameters.kt
```diff
@@ -90,15 +90,19 @@ class CertificateGenerationParameters : GenerationParameters {
...
-            keyUsage == that.keyUsage
+        return duration == that!!.duration &&
+            equalsIgnoringDuration(that)
+    }
+
+    fun equalsIgnoringDuration(other: CertificateGenerationParameters?): Boolean {
+        return keyLength == other!!.keyLength &&
+            isSelfSigned == other.isSelfSigned &&
+            isCa == other.isCa &&
+            (caName == other.caName || caName == null || other.caName == null) &&
```
bruce-ricard, line 101: I don't understand this line. Why isn't it just
```
caName == other.caName
```

like all other lines are tested?
If this logic is correct, I would like to somehow see some code that explains that logic, for example by extracting a method with a good name.

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
```diff
@@ -55,6 +55,9 @@ class CertificateCredentialValue : CredentialValue {
     @JsonIgnore
     var versionCreatedAt: Instant? = null
 
+    @JsonInclude(Include.NON_NULL)
+    var durationOverridden: Boolean? = null
```
bruce-ricard, line 59: Are you making this nullable for backwards compatibility with older certificates that won't have this property?

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
```diff
@@ -55,6 +55,9 @@ class CertificateCredentialValue : CredentialValue {
     @JsonIgnore
     var versionCreatedAt: Instant? = null
 
+    @JsonInclude(Include.NON_NULL)
```
bruce-ricard, line 58: I'm a bit confused by the `NON_NULL` tag together with the nullable value. Could you elaborate please?

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
```diff
@@ -113,4 +129,13 @@ private boolean shouldUseTransitionalParentToSign(final Boolean allowTransitiona
 
     return transitionalVersionCreatedAt.isAfter(latestNonTransitionalVersionCreatedAt);
   }
+
+  private int getCertificateDuration(final CertificateGenerationParameters params) {
+    if (params.isCa() && caMinimumDuration != null) {
```
bruce-ricard, line 134: Is there a reason you didn't make the `*MinimumDuration` variables default to `0` instead of `null`? 

That way, you wouldn't need to check if it's null, you just make the `max` call and it always works.

bruce-ricard (changes requested):
Added some inline comments and questions about the code.

Please let me know if something doesn't make sense.

Also, the first commit in the PR has its commit message starting with "WIP". Is that commit still a work in progress? If not could you amend it to get rid of the "WIP"?

applications/credhub-api/src/test/resources/application-minimum-duration.yml
```diff
@@ -0,0 +1,3 @@
+certificates:
+  ca_minimum_duration: 1825
```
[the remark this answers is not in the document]
peterhaochen47: I was less concerned about this, bc the bosh spec file would have [a description](https://github.com/pivotal/credhub-release/pull/61/files) where units can be specified (which is the existing patterns for all the other time-based config fields). And the existing CredHub API only accepts days as the unit for duration. Also,

backends/credhub/src/docs/asciidoc/snippets/introduction.adoc
peterhaochen47: Yeah this change is what I was [thinking about](https://github.com/pivotal/credhub-release/issues/60#issue-942313486) as well. However, Jesse's [initial proposal](https://github.com/pivotal/credhub-release/issues/60#issue-942313486) argues that accounting for duration in the `converge` mode (aka the existing behavior) is wrong to begin with. I personally don't have context on how consumers of credhub use the `converge` mode, but I believe `bosh` is the only credhub consumer that depends on this feature. So I'm kind of relying on the bosh team to judge whether the old behavior is a bug.

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
peterhaochen47: This is a [jackson annotation](https://www.baeldung.com/jackson-ignore-null-fields) that means: if this field has null value, do not include it when serializing this java object into json. Aka, ignore when null.

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/domain/CertificateGenerationParameters.kt
peterhaochen47: FYI this line is in the existing code. This PR only moves this line by extracting a method.

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
bruce-ricard: I don't understand what difference that makes then.

If you include null values, then the variable gets set to null.
If you don't, the variable stays set to null.

Doesn't it equal null regardless?

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/domain/CertificateGenerationParameters.kt
bruce-ricard: Ah, I see. I was wondering about that :).

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
peterhaochen47: This is just about how the resulting serialized json looks like. For example, when you have the java object `foo=abc, bar=null`:
- with the NON_NULL annotation, the resulting json looks like `{foo:abc}`
- without the NON_NULL annotation, the resulting json looks like `{foo:abc, bar:null}`

applications/credhub-api/src/test/resources/application-minimum-duration.yml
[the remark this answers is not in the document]
bruce-ricard: I don't like to rely on documentation for this. IMO it's much better when the variable name describes the unit. I understand that the existing pattern isn't to do that, but I'd argue that it's time for a change :).

applications/credhub-api/src/test/resources/application-minimum-duration.yml
```diff
@@ -0,0 +1,3 @@
...
+  leaf_minimum_duration: 1460
```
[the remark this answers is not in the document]
pvaramballypivot: Done.

applications/credhub-api/src/test/resources/application-minimum-duration.yml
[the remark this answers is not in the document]
pvaramballypivot: We updated both the config option name and variable naming for `ca_minimum_duration_in_days` and `leaf_minimum_duration_in_days`. We did not change the name of the pre-existing `duration` variable.

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
pvaramballypivot: Done.

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
ystros: While I'm generally in favor of better types, we don't think it adds a great deal of value in this case. The two "duration" fields are only used to overwrite the pre-existing `duration` field on the `CertificateGenerationParameters` class. That field comes in via the API and is already an integer in units of days:
https://github.com/cloudfoundry-incubator/credhub/blob/3a71bf0a2f00e8d031eb48569ddd083077ccbaf5/components/credentials/src/main/kotlin/org/cloudfoundry/credhub/domain/CertificateGenerationParameters.kt#L41

If we were to convert it to a Duration, we would just be immediately converting it back to an integer when we do set it on the generation parameters. See:
https://github.com/cloudfoundry-incubator/credhub/blob/65c2c17153bee5bb8ccfae307c83734e3000f116/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java#L133-L139
and
https://github.com/cloudfoundry-incubator/credhub/blob/65c2c17153bee5bb8ccfae307c83734e3000f116/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java#L61

The variables are not used anywhere else that would benefit from the more rich datatype. We've renamed the variables to include `InDays` to make units more clear. Let us know if that isn't sufficient and we can update it to be an actual `Duration` object.

backends/credhub/src/docs/asciidoc/snippets/introduction.adoc
ystros: As a member of the BOSH team, I think I would consider the old behavior a bug. The behavior of `converge` is meant to cause a credential to regenerate when the properties declared for the variable have changed. For things like a change to common name or signing CA, it makes a lot of sense that converge mode would cause the certificate to be regenerated - the underlying identity of the certificate has changed. This does not seem the case with duration / expiration date.

The cases this would be a breaking change would be as follows:
1. The BOSH operator has a manifest that defines a certificate using `update_mode: converge`.
2. They want to update *only* the duration field in order to either lengthen or shorten its expiration date.
3. They expect that doing a BOSH deploy will automatically regenerate the certificate with `update_mode: converge` for them using the new duration. The changes would cause it to not regenerate in this case.

I am not aware of any product or customer relying on this behavior. It seems very narrow to me, and can be solved by other workflows (such as making an explicit `credhub generate` CLI or equivalent API call). That said, if someone is relying on it, it would be a breaking change for them. 

We added an explicit feature flag on a temporary branch earlier that would default back to the earlier converge behavior unless turned on. It wasn't difficult, but we're holding off on merging it into this PR. Let us know which direction you would like to go in (release a new major, deem this as a bug fix and not worry about it, or add an explicit feature flag) and we'll proceed accordingly.

peterhaochen47: Thanks, @bruce-ricard and I can discuss tomorrow and get back to you.

Meanwhile @ystros, can I ask for your input? It seems like the tension is between:
1. Considering duration is a bug, because when the `converge` was introduced, it was never meant for only extending cert duration, which is a narrow and invalid use case. Hence, the lack of a feature flag does not constitute a breaking change, but a bug fix. CredHub can include the current PR in Credhub 2.10. BOSH can include this in a patch as well. 
2. Using the `converge` mode to only extend cert duration is a valid use case that people relies on. So CredHub should 1) include a feature flag (with the default being the existing behavior) and include this in a CredHub 2.10 or 2) bump major to CredHub 3.0 and BOSH would need to do the same.

What I'm trying to get at, is BOSH team might have better insight on whether using the `converge` mode to only extend cert duration is a valid use case that people relies on (CredHub, AFAIK, does not have a standalone cert use case outside of BOSH). Or to phrase the question a little differently: When the current PR (aka no feature flag) is merged and release, would the resulting new BOSH version be considered having a breaking change & needs major version bump? What would the BOSH team's decision be?

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
ystros: The general intent behind it being nullable (here and in other spots) is so that when the feature is disabled, the API does not return the field at all. This allows the corresponding CLI update (see https://github.com/cloudfoundry-incubator/credhub-cli/pull/110) to:

1. Not display `duration_overridden` in CLI output when the server-side feature is disabled
2. Not display `duration_overridden` in CLI output when talking to older versions of the CredHub server that doesn't have this feature (perhaps not a large problem)
3. Display `duration_overridden: true` in CLI output when the server-side feature is enabled and overridden.

It seemed heavy handed to always display `duration_overridden` in CLI output, so we arrived at making this nullable. However, your other feedback about this field never being set to `false` is a good catch.  We're considering adding a 4th case:
4. Display `duration_overridden: false` in CLI output when the server-side feature is enabled but the duration is NOT overridden.

backends/credhub/src/docs/asciidoc/snippets/introduction.adoc
bruce-ricard: OK right, fair enough. I actually talked with Jesse about that and I now remember suggesting that the current behavior is a bug...

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
ystros: We just pushed a change to the behavior here which I described in the other comment.
1. When the feature is not configured, the field is omitted (null)
2. When the feature is configured but the requested duration was used, the field is false.
3. When the feature is configured and the minimum duration was used instead, the field is true.

Let us know if this makes sense.

bruce-ricard:
You haven't replied or acted on the [comment](https://github.com/cloudfoundry-incubator/credhub/pull/201#pullrequestreview-730744205) about your first [commit](https://github.com/cloudfoundry-incubator/credhub/pull/201/commits/9fe6607b8bc59e28ead78364231ac7fbfe1b5eaf) starting with the letters "WIP". I am guessing that you missed it.

I still believe that it would be better to amend it. Let me know what you think.

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
bruce-ricard: OK thank you for the explanation, that makes a lot of sense.

There are 2 things I'd like to talk about: A general software point, and discussing these 4 options you mention.

I am a big believer in the [Software billion dollar mistake](https://hackernoon.com/null-the-billion-dollar-mistake-8t5z32d6). I believe that we should try to avoid using `null` as much as possible.
In this specific case, if you need a type containing 3 or 4 values, I would like to see some sort of enum created to represent all the possible options.

Regarding your specific question
>4. Display duration_overridden: false in CLI output when the server-side feature is enabled but the duration is NOT overridden.

I think that it does make sense.

I am also trying to think about the possibility of making the UX better. Putting myself in the shoes of a user, I am wondering how I feel about seeing a response containing `duration_overwritten: true`. "Overwritten to what?" I can hear myself think. Is the duration displayed in the CLI's output? If it is, then why bother add the "duration_overwritten" flag? The user will notice that it's not the one he asked for.

Completely different topic: forgive my non-native English, but I after some googling it seems to me like the proper spelling of the word is "overwritten". Google isn't too happy when you use "d"s instead of "t"s.

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
```diff
@@ -66,7 +69,8 @@ class CertificateCredentialValue : CredentialValue {
         certificateAuthority: Boolean,
         selfSigned: Boolean,
         generated: Boolean?,
```
bruce-ricard, line 71: I understand that this value is nullable. I dislike that very much. I very much would prefer this field to be a boolean. 

Developers often use boolean + null when they need a type with 3 values, I think that this is a mistake. We should created a type that clearly represents the 3 values that we want.

components/credentials/src/test/java/org/cloudfoundry/credhub/generators/CertificateGeneratorTest.java
```diff
@@ -317,6 +319,8 @@ public void whenSelfSignIsTrue_itGeneratesAValidSelfSignedCertificate() throws E
       equalTo(CertificateFormatter.pemOf(certificate)));
     assertThat(certificateCredential.getCa(), equalTo(CertificateFormatter.pemOf(certificate)));
     verify(signedCertificateGenerator, times(1)).getSelfSigned(rootCaKeyPair, inputParameters);
+    assertThat(certificateCredential.getDurationOverridden(),
+            equalTo(null));
```
bruce-ricard, line 323: When the `null` reaches the tests, we have gone too far.

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
bruce-ricard: I see. So I guess my main point, following the discussion about this field being nullable, is that if this field can't be null, we don't need any shady annotation.

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
```diff
@@ -113,4 +128,15 @@ private boolean shouldUseTransitionalParentToSign(final Boolean allowTransitiona
...
   }
+
+  private int getCertificateDuration(final CertificateGenerationParameters params) {
+    if (params.isCa()) {
+      return Math.max(params.getDuration(), caMinimumDurationInDays);
+    }
+    return Math.max(params.getDuration(), leafMinimumDurationInDays);
+  }
+
+  private boolean isMinimumDurationEnabled() {
```
bruce-ricard, line 139: That actually appears to be logically incorrect, isn't it? 

If the CA minimum duration is set to 3, and the leaf isn't set, and you are generating a leaf, this function will return `true`, which will result in the `durationOverwritten` field to be `true` or `false`, as if the feature was enabled for leaves. While it was only enabled for CAs.

So I didn't realize during my first review that it was important to know if the feature was enabled or not, for the purpose of showing the information in the response or not.

So it's annoying. It looks like you need to have 2 functions: `isMinimumDurationEnabledForCAs` and `...ForLeaves`.

Unless you actually assume that if a user sets one of them, then the whole feature is enabled. I guess that that's what you had in mind.

ystros:
Amended the initial commit to remove the "WIP".

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
ystros: We've updated the `duration_overridden` field to be non-nullable, so it will always appear regardless of whether the feature is enabled or not. That should make the API more consistent.

We have not renamed it - overridden is proper spelling, based off of the word override. Versus overwritten is based off of the word overwrote. Override seems more correct to us here, as the server is making a decision to reject the duration passed and use a different duration instead. But either word should work.

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
ystros: You may have meant to make this comment on a different line. The `generated` field existed before our PR. For historical purposes, I believe it was nullable because there was no way to know what the `true` or `false` value should be when it was originally added to the CredHub database. Perhaps an enum would have been better.

Regardless, we've updated the new `durationOverridden` field we've introduced to not be nullable. We did not touch the existing `generated` field.

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
ystros: > Unless you actually assume that if a user sets one of them, then the whole feature is enabled. I guess that that's what you had in mind.

Yes. We considered enabling either a leaf minimum or a CA minimum to be enabling the feature as a whole. However, since we've removed the null response, this method has been removed entirely.

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
ystros: Since the field is no longer nullable, it doesn't use this annotation anymore. 

However, since this class is used in contexts unrelated to the generate and regenerate endpoints, we have marked it instead with the `JsonIgnore` annotation. This directs the JSON renderer not to output the field at all. 

This was necessary since this class is also used in GET endpoints like `/api/v1/data?name=cert-name`, we do not want the field to appear at all - the duration being overridden (nor the originally requested duration) is not persisted data in the database per the requirements laid out in the original issue.

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
bruce-ricard: I did mean to comment on this line, and I did know that you did not write it :).

I meant to say that I realize that other fields are currently nullable, and I don't like that :)

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
bruce-ricard: >overridden is proper spelling, based off of the word override

Oh right, of course. I did some googling and I may have learned something about the different between "overwrite" and "override". Makes a lot of sense now, this appears to indeed be an override, and not an overwrite.

bruce-ricard: > We've updated the duration_overridden field to be non-nullable, so it will always appear in the API response regardless of whether the feature is enabled or not. That should make the API more consistent.

Now we lose the possibility to disable the feature for old users who don't need it and don't want breaking changes. This is important to us, so that we can ship this in a patch release,  and not have to maintain multiple versions (as mentioned in another conversation).

My previous comments were not talking at the feature level, only about the implementation detail of this field. I think that I was pretty happy with the feature.

Here is my position, let me know if you disagree:
1. I think that we need a type that allows to represent 3 values:
    1. Duration overridden (and obviously "feature enabled") -- the previous `true`
    2. Feature enabled but duration kept -- the previous `false`
    3. Feature disabled -- the previous `null`
1. I really want to avoid `null`, so I'd rather not use a nullable boolean.

Hence, I think that we need an Enum type to represent all these values, and depending on the values, we can decide what we return back.

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
bruce-ricard: I am confused again. Why did the "nullable" change impact the fact that the field should be included or not in JSON, and hence the `JsonIgnore` annotation?

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
bruce-ricard: 👌 

We may want this function back if we want to be able to disable the feature, and mentioned in a previous conversation.

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/CertificateGenerator.java
bruce-ricard: Hmm alright, indeed. I guess I was hoping that there was some intermediate layer in which we could have put the Duration. It looks kinda superfluous here. Although I think that I would have still done it myself, because I like high level types. Your call :).

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
ystros: We only wanted to conditionally include the field in responses for the generate / regenerate API endpoints. However, this class is used also with the GET endpoints. When the field was nullable, it only ever get set to a non-null value when generating or regenerating a certificate. Due to the previous `JsonInclude(Include.NON_NULL)` annotation, this meant it was not rendered for the GET endpoints, since the value was null on those code paths.

However, since it is now not nullable, it will always have a value (`false` for `duration_overridden`, `0` for `duration_used`). This means that the GET endpoints would start rendering:
```json
{ "name": "some-cert", ...more properties..., "duration_overridden": false, "duration_used": 0 }
```

which is not desirable - these fields only apply when generating/regenerating the certificate. So we added the `JsonIgnore` to ensure it is not output. This is now controlled by the new `CertificateGenerationView` class introduced in one of the latest commits. The existing `trustedCa` and `versionCreatedAt` fields on this `CertificateCredentialValue` class have similar `JsonIgnore` annotations, likely for the same reason.

peterhaochen47 (approved):
LGTM

backends/credhub/src/test/java/org/cloudfoundry/credhub/handlers/DefaultRegenerateHandlerTest.java
```diff
@@ -185,7 +185,7 @@ public void handleRegenerate_whenUserHasPermission_andAclsEnabled_regeneratesCre
       .thenReturn(savedCredentialVersion);
 
     CredentialView actualCredentialView = subjectWithAclsEnabled.handleRegenerate(CREDENTIAL_NAME, null);
-    CredentialView expectedCredentialView = CredentialView.fromEntity(savedCredentialVersion);
+    CredentialView expectedCredentialView = CredentialView.fromEntity(savedCredentialVersion, false, true);
```
bruce-ricard, line 188: very hard to understand what is happening in this function with the `false, true` parameters.

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/credential/CertificateCredentialValue.kt
```diff
@@ -55,6 +55,12 @@ class CertificateCredentialValue : CredentialValue {
...
 
+    @JsonIgnore
+    var durationOverridden: Boolean = false
+
+    @JsonIgnore
+    var durationUsed: Int = 0
+
     constructor() : super() {}
 
     constructor(
```
bruce-ricard, line 66: What a nightmare of a constructore this has become over time. I wanna try to refactor this.

bruce-ricard (changes requested):

PR #330, opened 2022-05-16
Add an option to start credhub jvm with debug agent loaded

hsinn0:
Made this work locally today and thought that it would be useful to have in repo


peterhaochen47:
LGTM, just tried it, and it worked.

PR #335, opened 2022-05-25
fix: Remove a pre-push hook routine that checks for `ssh-add -l` result

hsinn0:
- For most of us that are not using `ssh-add`, the pre-push hook routine fails
- Some of us work around the pre-push hook failure by doing `git push --no-verify`
- By having pre-push hook that actually works, developers can be encouraged to run the pre-push hooks when pushing to `main`

[#182032196]

### DIFF
--- a/hooks/pre-push
+++ b/hooks/pre-push
@@ -13,13 +13,6 @@ function go_to_project_root_directory() {
     cd "$hook_dir/.."
 }
 
-function check_ssh_key() {
-    if ! ssh-add -l >/dev/null; then
-        echo "No SSH key loaded! Please run vkl."
-        exit 1
-    fi
-}
-
 function run_linters() {
     ./scripts/lint.sh
 }
@@ -73,7 +66,6 @@ function main() {
     PID=
     set_bash_error_handling
     go_to_project_root_directory
-    check_ssh_key
 
     run_linters
     check_for_local_server


hooks/pre-push
```diff
@@ -13,13 +13,6 @@ function go_to_project_root_directory() {
     cd "$hook_dir/.."
 }
 
-function check_ssh_key() {
-    if ! ssh-add -l >/dev/null; then
-        echo "No SSH key loaded! Please run vkl."
-        exit 1
-    fi
-}
```
bruce-ricard, line 21: I'm trying to understand why this function was here in the first place.

I think that the reason is the following: 
This hook runs tests that take a while, and if they pass, it lets git run the `git push`. If you don't have an SSH key available for that, the `git push` will fail, and you will have lost one hour because you'll have to start again.

Because these days we develop on our personal workstations, we have a private key in ~/.ssh which will be used by git.

I've been looking into a couple of options, instead of just deleting the function.

1. We could keep the same message, but require an identity to be loaded in the ssh agent, but you could do it by simply running `ssh-add` instead of having to install another CLI like `vkl`
2. We could change the test to test for either an identity in the agent, or the presence of the file `~/.ssh/id_rsa` which will be used by git to push.
3. But the previous one actually makes me realize that even if an SSH identity is present, it still doesn't mean that it has permissions to actually push to the repo. So I was wondering if we could have a `git push --no-verify --dry-run` (or something like that) in the hook to make sure that you actually have permission to push.
4. But really all of this is making me move more and more towards the option of having pre-*commit*s hooks instead of pre-push ones. This way you can just make your commit when the tests pass, and you can push when you have permission to push.

hsinn0: Let's focus on the goal of this PR and its related story itself.  That was making the pre-hook funtions if one decided not to bypass via `git push --no-verify`.  I do not see any reason to keep the check_ssh_key now.  If you think that we need to have pre-commit hook, then bring it up in different places such as new story.

PR #378, opened 2022-09-12
Pin notAfter dates for certificates to max out at signing CA notAfter date

### LINKED ISSUE #377: Leaf certificate expiration reported by API does not account for CA expiration date
_What version of the credhub server you are using?_
2.12.8

_What version of the credhub cli you are using?_
2.6.2

_If you were attempting to accomplish a task, what was it you were attempting to do?_

As part of our CI/CD, we run `credhub curl -p /api/v1/certificates`, parse the output, and verify that the most recent version of all certificates has an `expiry_date` at least 30 days out. This helps give us plenty of time to initiate rotation procedures before expiration.

_What did you expect to happen?_

The `expiry_date` should be the value when the certificate ceases to function.

_What was the actual behavior?_

The `expiry_date` is parsed from the certificate data itself, without regard to the CA that signed it, which has a much closer in expiration date.

For example:

```bash
curl -p /api/v1/certificates
```

Output:

```json
{
  "certificates": [
    {
      "id": "xxx",
      "name": "/bosh/cf/diego_instance_identity_ca",
      "signed_by": "/bosh/cf/application_ca",
      "signs": [],
      "versions": [
        {
          "certificate_authority": true,
          "expiry_date": "2023-03-30T00:59:21Z",
          "generated": true,
          "id": "yyy",
          "self_signed": false,
          "transitional": false
        },
```

However a subsequent call to fetch more data about that version of the certificate:

```bash
credhub curl -p /api/v1/data/yyy | jq -r .value.ca | openssl x509 -noout -enddate
```

Reveals that the CA expires 6 months earlier (a few days ago):

```
notAfter=Sep  9 21:13:40 2022 GMT
```

We believe this is likely due to a previous rotation where update-mode converge may not have been correctly set, however none-the-less we think it is unhelpful for `credhub` to report an expiration date that is longer than the CA that it knows has signed it. Separately it would helpful to simply not issue a certificate longer than the parent CA as well.

_Please confirm where necessary:_

* [x] I have included a log output
* [ ] My log includes an error message
* [ ] I have included steps for reproduction

ae-govau:
This prevents one from issuing a certificate that appears to be valid, but in reality typically won't validate as being so.

Partially fixes #377

### DIFF
--- a/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/SignedCertificateGenerator.java
+++ b/components/credentials/src/main/java/org/cloudfoundry/credhub/generators/SignedCertificateGenerator.java
@@ -93,11 +93,24 @@ public class SignedCertificateGenerator {
 
     final BigInteger certificateSerialNumber = serialNumberGenerator.generate();
 
+    Date nvb = Date.from(now);
+    Data nva = Date.from(now.plus(Duration.ofDays(params.getDuration())));
+    if (issuerCertificate != null) {
+      // don't allow a child certificate to be valid before the signing certificate
+      if (nvb.isBefore(issuerCertificate.getNotBefore())) {
+        nvb = issuerCertificate.getNotBefore();
+      }
+      // don't allow a child certificate to be valid after the signing certificate
+      if (nva.isAfter(issuerCertificate.getNotAfter())) {
+        nva = issuerCertificate.getNotAfter();
+      }
+    }
+
     final JcaX509v3CertificateBuilder certificateBuilder = new JcaX509v3CertificateBuilder(
       issuerDn,
       certificateSerialNumber,
-      Date.from(now),
-      Date.from(now.plus(Duration.ofDays(params.getDuration()))),
+      nvb,
+      nva,
       params.getX500Principal(),
       keyPair.getPublic()
     );


ae-govau:
Accepting the CLA from our env is challenging (we've clicked through CLAs before for I think other parts of Cloudfoundry, but that particular one seems to be more complex for us to navigate for some reason). It's a trivial PR, feel free to pick up the few lines added into a new one if necessary.

Else, and if you'd like this change, let me know and I'll try harder to get the CLA organised.

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/SignedCertificateGenerator.java
```diff
@@ -93,11 +93,24 @@ private X509Certificate getSignedByIssuer(
 
     final BigInteger certificateSerialNumber = serialNumberGenerator.generate();
 
+    Date nvb = Date.from(now);
+    Data nva = Date.from(now.plus(Duration.ofDays(params.getDuration())));
```
swalchemist, line 97: This doesn't compile - should be "Date". Please run "gradlew test" after fixing.

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/SignedCertificateGenerator.java
```diff
@@ -93,11 +93,24 @@ private X509Certificate getSignedByIssuer(
 
     final BigInteger certificateSerialNumber = serialNumberGenerator.generate();
 
+    Date nvb = Date.from(now);
```
swalchemist, line 96: I suggest more descriptive names like notBefore and notAfter.

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/SignedCertificateGenerator.java
```diff
@@ -93,11 +93,24 @@ private X509Certificate getSignedByIssuer(
...
+    Data nva = Date.from(now.plus(Duration.ofDays(params.getDuration())));
+    if (issuerCertificate != null) {
```
swalchemist, line 98: Could you add unit test coverage for this new logic? I'm assuming we didn't already have good assertions for it, even though the tests hit this code.

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/SignedCertificateGenerator.java
```diff
@@ -93,11 +93,24 @@ private X509Certificate getSignedByIssuer(
...
+      // don't allow a child certificate to be valid before the signing certificate
+      if (nvb.isBefore(issuerCertificate.getNotBefore())) {
```
swalchemist, line 100: There isn't an `isBefore` method on Date, but you can use `before`.

components/credentials/src/main/java/org/cloudfoundry/credhub/generators/SignedCertificateGenerator.java
```diff
@@ -93,11 +93,24 @@ private X509Certificate getSignedByIssuer(
...
+        nvb = issuerCertificate.getNotBefore();
+      }
+      // don't allow a child certificate to be valid after the signing certificate
+      if (nva.isAfter(issuerCertificate.getNotAfter())) {
```
swalchemist, line 104: `isAfter` should be `after'.

swalchemist (changes requested):
Thanks, this looks useful, just needs to be tightened up. Also, first can you look into the EasyCLA issue and get that done? Let us know if you need some help with it.

Tallicia:
@ae-govau Do you plan to complete the CLA in the near future?

hsinn0:
Closing non active old PRs

PR #515, opened 2023-05-30
fix: update snakeyaml and ensure consistent subproject usage.

Tallicia:
Updating to latest snakeyaml without any vulnerabilities and ensuring subproject usage is consistent.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,6 +23,7 @@ buildscript {
         springSecurityOauth2Version = '2.5.2.RELEASE'
         springSecurityOauth2AutoconfigureVersion = '2.6.8'
         mariadbJdbcVersion = '2.7.9' // Bumping to v3 breaks some pipeline jobs, so pinning to v2 for now. v2 (current version) is stable and will be supported until about September 2025 (https://mariadb.com/kb/en/about-mariadb-connector-j/).
+        snakeyamlVersion = '2.0'
     }
     repositories {
         mavenCentral()
@@ -63,6 +64,9 @@ dependencyUpdates.resolutionStrategy = {
 subprojects {
     plugins.withType(JavaPlugin) {
         dependencies {
+            implementation("org.yaml:snakeyaml:${snakeyamlVersion}")
+            implementation("com.h2database:h2:${h2Version}")
+            implementation("com.google.guava:guava:${guavaVersion}")
             testImplementation("org.mockito:mockito-inline")
         }
     }


build.gradle
```diff
@@ -63,6 +64,9 @@ dependencyUpdates.resolutionStrategy = {
 subprojects {
     plugins.withType(JavaPlugin) {
         dependencies {
+            implementation("org.yaml:snakeyaml:${snakeyamlVersion}")
+            implementation("com.h2database:h2:${h2Version}")
+            implementation("com.google.guava:guava:${guavaVersion}")
```
bruce-ricard, line 69: Could you add in the commit message an explanation for these changes? Right now I have no idea why these changes are here.

build.gradle
```diff
@@ -23,6 +23,7 @@ buildscript {
         springSecurityOauth2Version = '2.5.2.RELEASE'
         springSecurityOauth2AutoconfigureVersion = '2.6.8'
         mariadbJdbcVersion = '2.7.9' // Bumping to v3 breaks some pipeline jobs, so pinning to v2 for now. v2 (current version) is stable and will be supported until about September 2025 (https://mariadb.com/kb/en/about-mariadb-connector-j/).
+        snakeyamlVersion = '2.0'
```
bruce-ricard, line 26: It looks like this is related to the "update snakeyaml" from the commit message.

Maybe "upgrade" would be more accurate? 

I don't understand why this is updating or upgrading the dependency. It looks like it was not pinned to anything before, which often means that we are always pulling the latest version. So yeah, I don't understand what is happening.

bruce-ricard (changes requested):
Please clarify the changes.

It looks like you may want to make 2 commits. The word "and" in the commit title is often a sign that your commit is doing too many things.

build.gradle
Tallicia: There are CVEs for snakeyaml that the default version of spring boot uses, this updates the usage of snakeyaml to be be an overridden version 2.0 which does not have any vulnerabilities. This is the update that has been applied in uaa 76 and 74 and is now being applied in credhub.

build.gradle
Tallicia: Yes, the commit message has been updated to include explanation.

build.gradle
Tallicia: @bruce-ricard  let me know if this is clear or if additional explanation is needed.

build.gradle
bruce-ricard: >The subprojects also need to consistently use guava, h2, and snakeyaml
to ensure all CVE resolutions are resolved throughout the
codebase.

OK. I understood that this was what you were doing. I don't understand why the h2 and guava bumps are related to the snakeyaml one. Is there a version dependency? Or are these just unrelated bumps that you happened to add to this commit?

build.gradle
Tallicia: I'll take thumbs up as this is clear. closing and merging, thank you!

Tallicia:
The commit is atomic to ensure the cve is fixed the and is to signifity both are requisite to address the cve and seperating would be insufficient.

build.gradle
peterhaochen47: I don't understand how the `h2database` is essential to the snakeyaml bump, or the CVE resolution.

peterhaochen47:
> The commit is atomic to ensure the cve is fixed the and is to signifity both are requisite to address the cve and seperating would be insufficient.
Why is the [h2database](https://github.com/cloudfoundry/credhub/blob/44ff4f4cc819db7b5f8892c7a2abcaed4c18a265/build.gradle#L68) specification to `1.4.200` requisite to address either the "snakeyaml 1.30 CVE" or "h2 1.4.200 CVE"?

PR #523, opened 2023-06-15
refactor: Override spring's kotlin version dependency once in the mai…

hsinn0:
…n build file instead of each subproject build file


Tallicia (approved):
This is much cleaner and easier to maintain, approved.

PR #538, opened 2023-07-06
refactor: Get fuller stacktrace from a flaky test

hsinn0:
- The test fails frequently in CI when running with postgresql with an exception.
- The stacktrace is chopped off as follows for example, so we cannot tell where exactly the exception is thrown in credhub code.
```
at org.springframework.dao.support.PersistenceExceptionTranslationInterceptor.invoke(PersistenceExceptionTranslationInterceptor.java:137)
... 11 more

```
- Increase the chance of getting fuller stacktrace by adding an ecxeption handler to print the exception cause's stacktrace
- Also remove unnecessary `throws` clauses in the method declarations.


hsinn0:
Closing this without merge as it does not really add any useful info in the test output

PR #586, opened 2023-09-27
fix: Incorrect not-rotated record count log message

hsinn0:
- As the rotated count can be greater than the starting count.
- Removed the log message.

[#185977500]

### DIFF
--- a/components/encryption/src/main/java/org/cloudfoundry/credhub/services/EncryptionKeyRotator.java
+++ b/components/encryption/src/main/java/org/cloudfoundry/credhub/services/EncryptionKeyRotator.java
@@ -59,8 +59,9 @@ public class EncryptionKeyRotator {
     if (rotatedRecordCount == 0) {
       LOGGER.info("Found no records in need of encryption key rotation.");
     } else {
-      LOGGER.info("Finished encryption key rotation in " + duration + " milliseconds. Details:");
-      LOGGER.info("  Successfully rotated " + rotatedRecordCount + " item(s)");
+      LOGGER.info("Finished encryption key rotation in " + duration +
+              " milliseconds. Details: Successfully rotated " +
+              rotatedRecordCount + " item(s)");
     }
 
     encryptionKeyCanaryMapper.delete(inactiveCanaries);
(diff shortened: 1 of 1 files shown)


components/encryption/src/main/java/org/cloudfoundry/credhub/services/EncryptionKeyRotator.java
```diff
@@ -38,9 +38,6 @@ public void rotate() {
     LOGGER.info("Starting encryption key rotation.");
     int rotatedRecordCount = 0;
 
-    final long startingNotRotatedRecordCount = encryptedValueDataService
-      .countAllByCanaryUuid(keySet.getActive().getUuid());
```
bruce-ricard, line 42: Could this `countAllByCanaryUuid(` method be removed from the required interface since we don't use it anymore?

bruce-ricard (changes requested):
Looks pretty good to me. 

If no tests are failing, we are close to be able to merge.

I just want to make sure that users have a way of validating that all entries have been rotated, especially when they have multiple CredHub instances. 
I feel like I would like, at the end of the rotation, a log of the type:
```
There are x remaining entries that are not encrypted with the newest key
```

hsinn0:
I am not sure how I also become a reviewer after adding my reply. Anyhow, I made another commit to combine two separate log message into one. Those should have been one log message to start with.

components/encryption/src/main/java/org/cloudfoundry/credhub/services/EncryptionKeyRotator.java
hsinn0: I cancelled my accidental comments as reviewer, and now lost all my previous comments made in response to @bruce-ricard's comments. So reconstructing the lost comments from my memory here:

>Could this countAllByCanaryUuid( method be removed from the required interface since we don't use it anymore?

`countAllByCanaryUuid()` is indeed not invoked by anything. Even so, I'd rather keep the it. The method does a specific database search. I don't see great benefit in removing it.

> I feel like I would like, at the end of the rotation, a log of the type:

Are you suggesting to make another database query to get the kind of data that we are getting rid of in this PR? In any case, I do not think that having that additional log message would be very useful. The code is executed by scheduled task and will be executed again next time the scheduled task kicks in.

Tallicia (approved):
This looks good to me for logn entry /removal clean up. Other issues can be addresses separately.

PR #599, opened 2023-10-09
Require Java 11 or later for all gradle tasks

hsinn0:
[#184142402]

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -73,3 +73,5 @@ subprojects {
 
 // Override spring boot's kotllin version dependency
 ext['kotlin.version'] = '${kotlinVersion}'
+
+assert JavaVersion.current().isJava11Compatible()


build.gradle
```diff
@@ -73,3 +73,5 @@ subprojects {
 
 // Override spring boot's kotllin version dependency
 ext['kotlin.version'] = '${kotlinVersion}'
```
swalchemist, line 75: Does the Kotlin version also need to be upgraded?

build.gradle
```diff
@@ -73,3 +73,5 @@ subprojects {
...
+
+assert JavaVersion.current().isJava11Compatible()
```
swalchemist, line 77: Hmmm, surprising that we didn't already having something like this for Java 8.

swalchemist (approved):

build.gradle
hsinn0: > Does the Kotlin version also need to be upgraded?

No, and we already tested the build with Java 11 and current Kotlin version.

build.gradle
hsinn0: The build probably had failed with JDK 1.7 anyhow, so no need to check, I guess. Right now, we can build the project with Java 8 just fine AFAICT. This assertion will block using JDK/JRE 8 for any gradle tasks, so prevents running into any possible untested issues later at runtime, etc.

PR #618, opened 2023-11-02
bump: Minimum Java version required for build

hsinn0:
- From 11 to 17

[#186054022]

_Approve but do not merge. I will merge the change along with other bump change later._

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -14,6 +14,6 @@ jobs:
         uses: actions/setup-java@v3
         with:
           distribution: 'liberica'
-          java-version: '11'
+          java-version: '17'
       - name: Build and test with Gradle
         run: ./gradlew clean build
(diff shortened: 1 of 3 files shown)


peterhaochen47 (approved):

peterhaochen47 (approved):

Tallicia:
It looks like there is a change in default precision causing the test failure: 

```
org.cloudfoundry.credhub.services.CertificateDataServiceTest > findAllValidMetadata when certificate is in 'names' returns certificates FAILED
    java.lang.AssertionError: expected:<2023-11-03T13:12:18.262579988Z> but was:<2023-11-03T13:12:18.262580Z>
```

hsinn0:
> It looks like there is a change in default precision causing the test failure:
> 
> 262579988Z != 262580Z

Hmm, I am not sure why I did not run into the same failure when I ran the test locally.

hsinn0:
> > It looks like there is a change in default precision causing the test failure:
> > 262579988Z != 262580Z
> 
> Hmm, I am not sure why I did not run into the same failure when I ran the test locally.

It is platform-dependent, e.g. Instant.now() return value precisions in my my Mac and linux are different.

Tallicia (approved):
LGTM

PR #639, opened 2023-11-28
Fix: confusing phrasing

bruce-ricard:
* "generates" isn't the correct word here. CredHub doesn't generate a key. It calculates (or computes) it.

* "deterministically" IMO can never be used with "generate", as "generate" implies "non deterministically", otherwise it's not generation, it's computation.

* "site-specific": I don't understand this term, or how it helps make the documentation clearer.

* "user-defined password": I think that "user-provided" is better.

* "a key": was changed multiple times to "the key". We are talking about _the_ key that will be used for encryption by CredHub, not any AES key.

* "lifetime of the server": this precison feels superfluous, and confusing. The lifetime of the CredHub server is so far about 7 years. I understand that the author meant "until the server instance stops running". I thought about rephrasing it, but since IMO it is superfluous, as this is exactly how memory works, I decided to remove it.

* I should have made multiple commits for this change. Apologies to micro-commit advocates.

### DIFF
--- a/docs/credhub-security-faq.md
+++ b/docs/credhub-security-faq.md
@@ -53,13 +53,17 @@ TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
 If CredHub's third-party integration with a Luna SafeNet HSM is
 enabled, then CredHub has no knowledge of key values.
 
-If CredHub is
-started in internal encryption mode, CredHub recalculates
-the AES256 key on startup. To do this, CredHub concatenates
-the user-provided encryption password from its configuration file with a
-randomly-generated salt stored in its database and hashes the
+If CredHub is started in internal encryption mode:
+
+* If a new encryption key is provided, CredHub will generate a new random salt string, and will store it in its database for the future.
+* If the encryption key had already been used before, CredHub will fetch the associated salt from its database to reuse it.
+
+CredHub then calculates the AES256 key by concatenating
+the user-provided encryption password from its configuration file with the salt, and hashes the
 resulting string. This hash is then used as the AES key, which CredHub
 will hold in memory.
+
+This design allows CredHub to have access to the AES key at all times, without ever having to store it on disk.
 
 ## How are privileged users prevented from compromising cryptographic
    keys?
--- a/docs/credhub-security-faq.md
+++ b/docs/credhub-security-faq.md
@@ -69,6 +69,12 @@ AES key, which CredHub will hold in memory.
 This design allows CredHub to have access to the AES key at all times,
 without ever having to store it on disk.
 
+Passwords and salts cannot be rotated individually, and they are never
+automatically rotated. If you want to rotate either or both of them,
+you need to add a new key to CredHub's configuration, and CredHub will
+start using that new password and a newly generated salt to create a
+fresh AES key.
+
 ## How are privileged users prevented from compromising cryptographic
    keys?
 
--- a/docs/credhub-security-faq.md
+++ b/docs/credhub-security-faq.md
@@ -52,12 +52,12 @@ TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
 
 If CredHub's third-party integration with a Luna SafeNet HSM is
 enabled, then CredHub has no knowledge of key values. If CredHub is
-started in internal encryption mode, CredHub deterministically
-generates a site-specific AES256 key on startup. CredHub concatenates
-a user-defined password from its configuration file with a
+started in internal encryption mode, CredHub recalculates
+the AES256 key on startup. To do this, CredHub concatenates
+the user-provided encryption password from its configuration file with a
 randomly-generated salt stored in its database and hashes the
-resulting string to deterministically generate an AES256 key, which it
-holds in memory for the lifetime of the server.
+resulting string. This hash is then used as the AES key, which CredHub
+will hold in memory.
 
 ## How are privileged users prevented from compromising cryptographic
    keys?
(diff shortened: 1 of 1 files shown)


docs/credhub-security-faq.md
```diff
@@ -52,12 +52,12 @@ TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
 
 If CredHub's third-party integration with a Luna SafeNet HSM is
 enabled, then CredHub has no knowledge of key values. If CredHub is
-started in internal encryption mode, CredHub deterministically
-generates a site-specific AES256 key on startup. CredHub concatenates
-a user-defined password from its configuration file with a
+started in internal encryption mode, CredHub recalculates
```
bruce-ricard, line 55: Maybe "calculates" is better? It could be the first time it calculates it, if the user just provided a new encryption password. But it will **re**calculate it every time after that.

How can we phrase this?

swalchemist: I'm having trouble understanding what key it's talking about here. I was assuming it's the key used to encrypt data at rest, but I see that it's changed on startup, which doesn't fit that idea. Is this key used to communicate with clients?

bruce-ricard: Perfect, thanks for sharing that. That is pretty much exactly the confusion I'm trying to solve. Apparently my first attempt is a failure.

It *is* the key used to encrypt data at rest.

> I see that it's changed on startup

Which part makes you say that? Because it's not.

bruce-ricard: The previous version of the docs says that it "generates a site-specific AES256 key on startup" which IMO is very confusing, as it appears that it creates a new key. But it doesn't. It's the same key every time that is computed. That's how it works.

swalchemist: Ah, ok. So saying "recalculates" I think is just as confusing as "generates". I don't understand what you're distinguishing between the first startup and subsequent startups, if the action is the same and the resulting key is the same each time.

docs/credhub-security-faq.md
```diff
@@ -52,12 +52,12 @@ TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
...
+the AES256 key on startup. To do this, CredHub concatenates
+the user-provided encryption password from its configuration file with a
 randomly-generated salt stored in its database and hashes the
```
swalchemist, line 58: It's not clear from this when CredHub is randomly generating the salt. Could be at every startup from what this says.

docs/credhub-security-faq.md
bruce-ricard: > the action is the same and the resulting key is the same each time

Yes

> I don't understand what you're distinguishing between the first startup and subsequent startups

Fair enough. So would "calculates" be better then?

docs/credhub-security-faq.md
bruce-ricard: OK. To me, "stored in its database" was enough to explain that it is only generated the first time.

I'm going to rewrite the whole thing, and split it in 2 sections:
1. what happens the first time
2. what happens on the next restarts of the server

that should make everything clearer.

swalchemist: For what it's worth, the other edits look fine.

docs/credhub-security-faq.md
swalchemist: What made me think it was changed at every startup was "recalculates" and the "randomly generated" part of the salt (as a naive reader). Now that I understand better, I agree that "generate" isn't a good way to explain it. So, yes, I like "calculates". And in the spirit of the "deterministically" part that you removed, maybe a brief explanation like "The key will be the same every time CredHub starts as long as the user-provided password has not changed."

bruce-ricard:
Something's wrong with my commit signatures. I have to fix that.

docs/credhub-security-faq.md
```diff
@@ -51,13 +51,29 @@ TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
...
-generates a site-specific AES256 key on startup. CredHub concatenates
-a user-defined password from its configuration file with a
-randomly-generated salt stored in its database and hashes the
-resulting string to deterministically generate an AES256 key, which it
-holds in memory for the lifetime of the server.
+enabled, then CredHub has no knowledge of key values.
+
+If CredHub is started in internal encryption mode:
+
+* If a new encryption key is provided, CredHub will generate a new
```
swalchemist, line 58: Should this say "If a new encryption password is provided"? Is it a password or a key? Same for the bullet below.

docs/credhub-security-faq.md
```diff
@@ -51,13 +51,29 @@ TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
...
+
+CredHub then calculates the AES256 key by concatenating the
+user-provided encryption password from its configuration file with the
+salt, and hashes the resulting string. This hash is then used as the
+AES key, which CredHub will hold in memory.
+
+This design allows CredHub to have access to the AES key at all times,
+without ever having to store it on disk.
+
+Passwords and salts cannot be rotated individually, and they are never
```
swalchemist, line 72: This could probably be simplified. What people most likely care about is the the salt that can't be rotated by itself. This already explains that the salt will be changed when the password is changed. Maybe -

```
The password and salt are never automatically rotated. Also, the salt cannot be rotated by itself without changing the password.
```

swalchemist (changes requested):

swalchemist (approved):

PR #679, opened 2024-01-13
Code quality improvements inspired by gradle check/shellcheck/sonarlint

swalchemist:
* Fix several warnings that show up in the credhub-edge "lint" job (from "gradle check"). This was pretty much all Kotlin code.
* Additional refactoring for consistency.
* Some small README file improvements.
* Broadens the scope of shellcheck to include the 2 docker scripts, plus fixing the problems that this finds.
* Fix a few additional issues found by "shellcheck --enable=all". 
* Fix sonarlint warnings in the Dockerfile.

### DIFF
--- a/docker/setup_trust_store.sh
+++ b/docker/setup_trust_store.sh
@@ -19,11 +19,11 @@ generate_server_ca() {
 }
 
 setup_tls_key_store() {
-    if [ -f "$SERVER_CA_CERT_PATH" ] && [ -f "$SERVER_CA_PRIVATE_KEY_PATH" ]
+    if [[ -f "${SERVER_CA_CERT_PATH}" && -f "${SERVER_CA_PRIVATE_KEY_PATH}" ]]
     then
       echo "Using provided server CA"
-      cp "$SERVER_CA_CERT_PATH" server_ca_cert.pem
-      cp "$SERVER_CA_PRIVATE_KEY_PATH" server_ca_private.pem
+      cp "${SERVER_CA_CERT_PATH}" server_ca_cert.pem
+      cp "${SERVER_CA_PRIVATE_KEY_PATH}" server_ca_private.pem
     else
       generate_server_ca
     fi
@@ -31,7 +31,7 @@ setup_tls_key_store() {
     cat > server.cnf <<EOF
 [v3_ca]
 subjectKeyIdentifier=hash
-subjectAltName=$SUBJECT_ALTERNATIVE_NAMES
+subjectAltName="${SUBJECT_ALTERNATIVE_NAMES}"
 EOF
 
     echo "Generating a key store for the certificate the server presents during TLS"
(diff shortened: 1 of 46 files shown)


docker/setup_trust_store.sh
```diff
@@ -64,9 +64,9 @@ setup_auth_server_trust_store() {
     -trustcacerts \
     -noprompt \
     -alias "uaa_ca" \
-    -file "${UAA_CA_PATH}" \
+    -file "${UAA_CA_PATH:?}" \
```
bruce-ricard, line 67: Why are you adding `:?` to all variable lookups? It seems like it has the exact same effect as `set -u` (`--nounset`), which is present at the beginning of the script.

scripts/lint.sh
```diff
@@ -14,7 +14,7 @@ function lint_scripts() {
 
 function lint_jvm_language_code() {
     ./gradlew ktlintFormat
-    ./gradlew check -x test
+    ./gradlew clean check -x test
```
bruce-ricard, line 17: This forces an entire rebuild though. It's going to waste a lot of time.

It feels like there should be a better way to achieve this.

bruce-ricard (changes requested):
Thanks Danny. Almost good to go.

docker/setup_trust_store.sh
swalchemist: It is redundant with `set -u`, except that shellcheck wants some reassurance that those variables are all initialized already and doesn't accept `set -u` as a sufficient solution.

I would have preferred to use a declare statement instead, basically declaring at the top of the script what variables it's depending on being already set in the environment. But that also has the side effect of giving them a default value and circumvents the benefit of the `set -u`.

Do you have a better suggestion for making shellcheck happy in the case where a variable is not initialized in the script where it's referenced?

scripts/lint.sh
swalchemist: If you want to check all the code, I think you have to build all of it.

The current script before this fix is broken, because it only checks the code that hasn't already been built locally.

docker/setup_trust_store.sh
swalchemist: It would a better bug report if you had used "--enable=check-unassigned-uppercase" instead of "--enable=all". It looks like a doc problem to me.

scripts/lint.sh
swalchemist: Where did we build everything before the "gradle clean check"?

docker/setup_trust_store.sh
swalchemist: I have removed all the added ":?" notations so we can get this PR merged.

scripts/lint.sh
bruce-ricard: Well, whenever we do what the "gradle clean" cleans.

bruce-ricard (approved):
Let's ship it.

We can always revert the "clean" change if it causes problems in the future.

scripts/lint.sh
swalchemist: That could have been any other build activity unrelated to the current gradle check run. And that's why the results were inconsistent.

PR #682, opened 2024-01-20
Sonar lint fixes

swalchemist:
Mitigates a handful of SonarLint findings (codes for each are given in the commit messages). Also includes a new test to cover some of the modified code, and some related refactoring.

There are also a large number of changes to migrate some of the tests to JUnit 5 and improve how they assert on throwing an exception. Note that the `@Test(expected = ...` instances were migrated because this was required to use JUnit 5, but the JUnit 3 style exception assertions that use try/catch are not yet migrated, in order to reduce the volume of changes in this PR.

This was a draft PR until #679 was merged first.


components/credentials/src/test/java/org/cloudfoundry/credhub/requests/CertificateGenerationRequestParametersTest.java
```diff
@@ -17,6 +19,9 @@ public class CertificateGenerationRequestParametersTest {
 
   private CertificateGenerationRequestParameters subject;
 
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
```
bruce-ricard: So you decided not to go for the lambda way?

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/requests/CertificateGenerationRequestParameters.kt
```diff
@@ -84,9 +84,13 @@ class CertificateGenerationRequestParameters {
             isEmpty(country)
         ) {
             throw ParameterizedValidationException(ErrorMessages.MISSING_CERTIFICATE_PARAMETERS)
-        } else if (isEmpty(caName) && !selfSigned && !isCa) {
+        }
+
+        if (isEmpty(caName) && !selfSigned && !isCa) {
```
bruce-ricard, line 95: This change makes the code harder to understand. Now we have to think about which conditional block we are entering. Before, it was clear from the structure that we can only enter 1.

I disagree with this. I don't know what "when" is in kotlin, I guess it would be better than this.

components/errors/src/main/kotlin/org/cloudfoundry/credhub/exceptions/InvalidAlternateNameCertificateException.kt
```diff
@@ -0,0 +1,8 @@
+package org.cloudfoundry.credhub.exceptions
+
+import org.cloudfoundry.credhub.ErrorMessages
+
+class InvalidAlternateNameCertificateException @JvmOverloads
+    constructor(messageCode: String = ErrorMessages.INVALID_ALTERNATE_NAME, parameters: Array<Any> = arrayOf())
```
bruce-ricard: It feels like now the error message constant could and should be moved to this exception class.

bruce-ricard (changes requested):
Looks very good. I just have an issue with the conditional block change.

It looks like my comment about the exception checking is mute, as a later commit appears to refactor them.

components/credentials/src/test/java/org/cloudfoundry/credhub/requests/CertificateGenerationRequestParametersTest.java
swalchemist: Oh, I missed that one. I'll change it.

components/credentials/src/main/kotlin/org/cloudfoundry/credhub/requests/CertificateGenerationRequestParameters.kt
swalchemist: I found this easier to understand than before, because there are also several independent conditionals below the if/else if/if structure that I changed. Now it's a series of conditionals.

Is it not clear that these are throwing exceptions?

components/credentials/build.gradle
```diff
@@ -44,8 +44,8 @@ dependencies {
     implementation('org.springframework.boot:spring-boot-starter-security')
     implementation('org.springframework.boot:spring-boot-starter-validation')
     testImplementation("org.springframework.boot:spring-boot-starter-test")
-    testImplementation("junit:junit")
-
+    testImplementation(platform('org.junit:junit-bom:5.10.2'))
```
hsinn0, line 47: Specifying version here is unnecessary. I will commit a change to address this myself.

hsinn0 (approved):

Issue #689, opened 2024-02-05
CredHub 2.12.59+ docker images do not run successfully


peterhaochen47:
What version of the credhub server you are using?
2.12.59+

What version of the credhub cli you are using?
Any

If you were attempting to accomplish a task, what was it you were attempting to do?
Starting credhub server using its docker image

What did you expect to happen?
Below are the successful steps:
```
# check out the last commit before credhub 2.12.59
git checkout d2b7d8df0544b73972075fe0f6add5ff8718e7ab 
# build the credhub docker image and record its docker image id
docker build .
# run the image and see all the server start logs
docker run --publish 9000:9000 [DOCKER_IMAGE_ID]
# can login to the started credhub server and run credhub-cli commands
credhub login -s https://localhost:9000 --client-name=credhub_client --client-secret=secret --skip-tls-validation
```

What was the actual behavior?
Below are the steps of reproduction:
```
# check out any commit in credhub 2.12.59+ (e.g. the current top of the main branch)
git checkout 4b8fca5dcf35adb68cab7dc1ad4ebe43ca8af0c0 
# build the credhub docker image and record its docker image id
docker build .
# run the image and DO NOT see all the server start logs
docker run --publish 9000:9000 [DOCKER_IMAGE_ID]
# CANNOT login to the started credhub server and run credhub-cli commands
credhub login -s https://localhost:9000 --client-name=credhub_client --client-secret=secret --skip-tls-validation
```

Please confirm where necessary:
* [ ] I have included a log output
* [ ] My log includes an error message
* [x] I have included steps for reproduction

## notes
* Slack thread where this problem is initially reported: https://cloudfoundry.slack.com/archives/C3EN0BFC0/p1706791893295509
* commits included in CredHub 2.12.59: https://github.com/cloudfoundry/credhub/compare/c0ef6452f5081fc75e70e9f8b401062d09378c68...7c2b251964db12d13b039e797aba170f210880f2

silvestre:
@peterhaochen47 : Based on the steps you provided I was able to use `git bisect run` using the following script:
```
#!/bin/bash

set -euxo pipefail

docker build . -t credhub-server
# stop the server from a previous run
docker stop credhub-server || true
# wait for the server to stop
while docker container inspect credhub-server >/dev/null 2>&1; do sleep 1; done
docker run --detach --rm --name credhub-server --publish 9000:9000 credhub-server
# wait for the server to start - a health check would be better
sleep 20
docker logs credhub-server || true
[...]

PR #690, opened 2024-02-06
fix(docker): fix docker image startup

silvestre:
Fixes: #689


swalchemist:
Thanks! We'll take a look.

peterhaochen47:
Verified that this works. Thanks!

Issue #190, opened 2021-05-07
deletion of old credential versions not possible


dueckminor:
What version of the credhub server you are using?
2.9.0

What version of the credhub cli you are using?
2.9.0

If you were attempting to accomplish a task, what was it you were attempting to do?
I rotate/update (non-certificate) credentials regularly using the credhub cli by either using:
- `credhub set`
- `credhub generate`
- `credhub regenerate`

What did you expect to happen?
I would expect that this doesn't decrease the overall CredHub performance and that the CredHub CLI/API would allow me to delete old versions of the credential. For certificate credentials, there is such an API, but not for all other types of credentials.

What was the actual behaviour?
All currently available method to update a credential will lead to an addition version of the same credential. There is currently no API available which allows to remove old and obsolete versions of one credential.
If the number of versions grows (more than 5000), the performance of CredHub dramatically decreases and CredHub finally gets unhealthy (I observed a high CPU load on the Database)

Please confirm where necessary:
* [ ] I have included a log output
* [ ] My log includes an error message
* [ ] I have included steps for reproduction

If you are a PCF customer with an Operation Manager (PCF Ops Manager) please direct your questions to support (https://support.pivotal.io/)

bruce-ricard:
I guess deleting the credential and recreating it could be a workaround. If you want to cleanup all old versions of a credential, you can run `credhub delete --name ...` and then `credhub generate` or `set` again. All older versions will be gone.

But I agree that this is a pretty valid feature request.

bruce-ricard:
Sorry about the wrong link just above. The PR fixes another issue.

PR #701, opened 2024-03-06
fix: Orphaned encrypted_value records

hsinn0:
- When a credential was deleted, its related encrypted_value records were not being deleted.
- Implemented a trigger to make sure that those records are deleted.

[#182121168]

There will be more commits, including tests, triggers for other databases and indexes to address delete operation performance.

### DIFF
--- a/applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
+++ b/applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
@@ -7,44 +7,53 @@ DELIMITER |
 CREATE TRIGGER tr_credential_deleted BEFORE DELETE ON credential
 FOR EACH ROW
 BEGIN
-  DROP TEMPORARY TABLE IF EXISTS
-    tmp_credential_version_uuids, tmp_encrypted_value_uuids;
+    DECLARE done INT DEFAULT FALSE;
+    DECLARE cred_ver_id, enc_val_id BINARY(16);
+    DECLARE cred_ver_type VARCHAR(31);
+    DECLARE cur CURSOR FOR
+        SELECT type, uuid FROM credential_version
+            WHERE credential_uuid = OLD.uuid;
+    DECLARE CONTINUE HANDLER FOR NOT FOUND SET done = TRUE;
 
-  -- Remember the uuids of the credential_version records being deleted
-  CREATE TEMPORARY TABLE tmp_credential_version_uuids
-    SELECT uuid FROM credential_version
-      WHERE credential_uuid = OLD.uuid;
+    OPEN cur;
+    read_loop: LOOP
+        FETCH cur into cred_ver_type, cred_ver_id;
+        IF done THEN
+          LEAVE read_loop;
+        END IF;
 
-  -- Remember the uuids of the encrypted_value records to delete
-  CREATE TEMPORARY TABLE tmp_encrypted_value_uuids
-    SELECT encrypted_value_uuid FROM credential_version
-      WHERE credential_uuid = OLD.uuid;
-  INSERT INTO tmp_encrypted_value_uuids (encrypted_value_uuid)
-    SELECT password_parameters_uuid from password_credential
-      WHERE password_credential.uuid IN
-        (SELECT uuid FROM tmp_credential_version_uuids);
-  INSERT INTO tmp_encrypted_value_uuids (encrypted_value_uuid)
-    SELECT password_parameters_uuid from user_credential
-      WHERE user_credential.uuid IN
-        (SELECT uuid FROM tmp_credential_version_uuids);
+        -- For each talbe record that is referencing encrypted_value record,
+        -- set FKs to encrypted_value to null before deleting the encrypted_value
+        -- record. Otherwise, the referencing record cannot be deleted because of
+        -- the FK constraint.
 
-  -- Set FKs to encrypted_value to null so the encrypted_value
-  -- record can be deleted without violating the FK constraint
-  UPDATE credential_version SET encrypted_value_uuid = NULL
-    WHERE credential_uuid = OLD.uuid;
-  UPDATE password_credential SET password_parameters_uuid = NULL
-    WHERE password_credential.uuid IN
-      (SELECT uuid FROM tmp_credential_version_uuids);
-  UPDATE user_credential SET password_parameters_uuid = NULL
-    WHERE user_credential.uuid IN
-      (SELECT uuid FROM tmp_credential_version_uuids);
+        -- Delete the encrypted_value record that is associated with the
+        -- credential_version record.
+        SELECT encrypted_value_uuid FROM credential_version
+            WHERE uuid = cred_ver_id INTO enc_val_id;
+        UPDATE credential_version SET encrypted_value_uuid = NULL
+            WHERE uuid = cred_ver_id;
+        DELETE FROM encrypted_value WHERE uuid = enc_val_id;
+        SET enc_val_id = NULL;
 
-  -- Delete the encrypted_value records
-  DELETE FROM encrypted_value WHERE encrypted_value.uuid IN
-    (SELECT encrypted_value_uuid from tmp_encrypted_value_uuids);
-
-  DROP TEMPORARY TABLE IF EXISTS
-    tmp_credential_version_uuids, tmp_encrypted_value_uuids;
+        -- For password or user type credential, the password_credential
+        -- or user_credential record also has associated encrypted_value record.
+        -- Delete them, too.
+        IF cred_ver_type = 'password' THEN
+            SELECT password_parameters_uuid from password_credential
+                WHERE uuid = cred_ver_id INTO enc_val_id;
+            UPDATE password_credential SET password_parameters_uuid = NULL
+                WHERE uuid = cred_ver_id;
+            DELETE FROM encrypted_value WHERE uuid = enc_val_id;
+        ELSEIF cred_ver_type = 'user' THEN
+            SELECT password_parameters_uuid from user_credential
+                WHERE uuid = cred_ver_id INTO enc_val_id;
+            UPDATE user_credential SET password_parameters_uuid = NULL
+                WHERE uuid = cred_ver_id;
+            DELETE FROM encrypted_value WHERE uuid = enc_val_id;
+        END IF;
+    END LOOP read_loop;
+    CLOSE cur;
 END;
 |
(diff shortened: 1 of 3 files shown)


applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
```diff
@@ -0,0 +1,50 @@
+-- Fix for https://github.com/cloudfoundry/credhub/issues/231
+-- Create delete trigger on credential table instead of the tables directly
```
swalchemist, line 2: It's not quite clear when I first read this which tables "instead of on the tables directly" refers to.

hsinn0: I think that the SQL code itself shows what tables we are talking about here. Do you think that we should list table names in the comment? BTW, it has more comments that most of the existing flyway migration scripts. We usually do not have any comments.

applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
```diff
@@ -0,0 +1,50 @@
...
+-- reference encrypted_value table because the trigger does not work for
+-- cascade-deleted records in MySQL.
+DROP TRIGGER IF EXISTS tr_credential_deleted;
+DELIMITER |
+CREATE TRIGGER tr_credential_deleted BEFORE DELETE ON credential
+FOR EACH ROW
+BEGIN
+  DROP TEMPORARY TABLE IF EXISTS
+    tmp_credential_version_uuids, tmp_encrypted_value_uuids;
```
bruce-ricard, line 11: Are we sure that this can't delete the tables of another delete trigger running in parallel?

applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
```diff
@@ -0,0 +1,50 @@
...
+
+  -- Remember the uuids of the credential_version records being deleted
+  CREATE TEMPORARY TABLE tmp_credential_version_uuids
+    SELECT uuid FROM credential_version
+      WHERE credential_uuid = OLD.uuid;
```
bruce-ricard, line 16: I was hoping this could be done without creating temporary tables, this looks pretty heavy weight for what we are trying to do.

Is it really our only option to create 2 temporary tables to store one ID in each?

applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
```diff
@@ -0,0 +1,50 @@
...
+    WHERE credential_uuid = OLD.uuid;
+  UPDATE password_credential SET password_parameters_uuid = NULL
+    WHERE password_credential.uuid IN
+      (SELECT uuid FROM tmp_credential_version_uuids);
+  UPDATE user_credential SET password_parameters_uuid = NULL
+    WHERE user_credential.uuid IN
+      (SELECT uuid FROM tmp_credential_version_uuids);
+
+  -- Delete the encrypted_value records
+  DELETE FROM encrypted_value WHERE encrypted_value.uuid IN
```
Tallicia, line 43: Why is a temporary table needed instead of a nested subquery?

Additionally, what is the FK providing since this is bypassing it?

Finally, This appears to be more complex and creating a higher maintenance cost than utilizing then ORMs orphanRemoval capability, how is this being compared to that approach?

components/credentials/src/test/java/org/cloudfoundry/credhub/services/DefaultCredentialVersionDataServiceTest.java
```diff
@@ -272,6 +277,7 @@ public void delete_onAnExistingCredential_returnsTrue() {
 
   @Test
   public void delete_onACredentialName_deletesAllCredentialsWithTheName() {
+    long nEncryptedValuesPre = encryptedValueRepository.count();
```
bruce-ricard, line 280: Could this variable have a better name, please?

applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
hsinn0: I will rephrase my argument. The comment combined with the code itself is more than sufficient.

applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
hsinn0: Only if you have another trigger that uses the same tables, which we do not.

applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
hsinn0: Temporary tables are in-memory tables and not too heavy weight. That is only proper way to that can be done in MySQL that I know of.

applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
hsinn0: > Why is a temporary table needed instead of a nested subquery?

Mainly because the values need to be retrieved before running the next query.

> Additionally, what is the FK providing since this is bypassing it?
Not sure what you mean. I do not think that you understood the context. The trigger deletes the records that cannot be handled by FK cascade delete because the FK direction is the other way.

> This appears to be more complex and creating a higher maintenance cost
There is no higher  maintenance cost. The script is written once and done.

> than utilizing then ORMs orphanRemoval capability, how is this being compared to that approach
My understanding is that more than a week was spent in that effort without producing any useful result. Based on what I see, going that route required very involved Java/Kotlin code changes. This requires no Java/Kotlin code change.

components/credentials/src/test/java/org/cloudfoundry/credhub/services/DefaultCredentialVersionDataServiceTest.java
```diff
@@ -301,10 +307,13 @@ public void delete_onACredentialName_deletesAllCredentialsWithTheName() {
 
     assertThat(subject.findAllByName("/my-credential"), hasSize(0));
     assertNull(credentialDataService.find("/my-credential"));
+    assertEquals("Associated encryptedValues were deleted",
+            nEncryptedValuesPre, encryptedValueRepository.count());
```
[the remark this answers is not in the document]
hsinn0: As I said before, cleanup verification should be part of deletion test. Additional benefit is that this is lesser coding while accomplish the same coverage. BTW. I need to add more tests for different credential types.

components/credentials/src/test/java/org/cloudfoundry/credhub/services/DefaultCredentialVersionDataServiceTest.java
hsinn0: nSomething is a common variable name style for number of something. I like it better than somethingCount, countOfSomething, or numberOfSomething.

applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
hsinn0: Temp table use could be avoided if we delete encrypted_value records that are not referenced by those 3 tables anymore after the those 3 table records were deleted. However, the performance will be worse that way because you now have to do full scan of the 3 tables every time.

applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
hsinn0: I have confirmed that the MySQL temp table scope is per connection. So doing `credhub delete` from different client for example, won't be a problem. However, there is possibility that the process is triggered multiple times in the same connection as it is possible that multiple delete operations can be done in the same connection. I will re-work the trigger code to protect from that scenario.

applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
hsinn0: I am going to rewrite it not to use temp tables. There are potential issues with using temp tables inside trigger as we do not have complete control of how/when the trigger is executed.

applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
hsinn0: I am going to rewrite it not to use temp tables. There are potential issues with using temp tables inside trigger as we do not have complete control of how/when the trigger is executed.

hsinn0:
Rewrote mysql trigger to use cursor instead of temp table. Also got rid of previous commit with test code change as the added code to the existing test cases had issues giving inconsistent result. Will add new test cases instead.

applications/credhub-api/src/main/resources/db/migration/mysql/V57_2__delete_related_encrypted_value_when_credential_is_deleted.sql
hsinn0: The trigger had been changed not to use temp tables.

bruce-ricard (approved):
Amazing work

bruce-ricard:
This PR attempts to resolve https://github.com/cloudfoundry/credhub/issues/190

PR #718, opened 2024-04-03
doc: The need to stay with protobuf 3.25.x

hsinn0:
Documenting instead of adding to dependabot ignore list. Will have to disregard 4.x `protobuf` dependabot PRs for a while, but the expectation is that `grpc` will be updated  to bump its dependency to `protobuf` in near future, hopefully.

### DIFF
--- a/backends/credhub/build.gradle
+++ b/backends/credhub/build.gradle
@@ -1,7 +1,4 @@
 buildscript {
-    ext {
-        grpcVersion = '1.62.2'
-    }
     repositories {
         mavenCentral()
         maven { url("https://repo.spring.io/plugins-release") }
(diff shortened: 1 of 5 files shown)


Tallicia (approved):
This is good. Monthly checks for updates should keep this updated if there are any more 3.xx updates, yes? Or is 3.xx no longer supported?

peterhaochen47 (changes requested):
Could we maybe use a variable, similar to `commonsCodecVersion`, that is declared in the [main build.gradle](https://github.com/cloudfoundry/credhub/blob/3d20de779bc62f11a5a073d484cb9272eff07b9c/build.gradle) where the comment can also go. This way, all the intentional version pinning and their justifications/criterion for unpinning are all in the same place.

hsinn0:
> This is good. Monthly checks for updates should keep this updated if there are any more 3.xx updates, yes? Or is 3.xx no longer supported?

3.xx is not end of support. We could configure dependabot to ignore major version bumps only. However for this, I'd like to see all version bump PRs because its relationship with grpc version, as explained in the comment.

peterhaochen47 (approved):

PR #720, opened 2024-04-04
fix: remove triggers causing failed startup

Tallicia:
Updating migrations to remove triggers which can cause startup failures and delete triggers to remain consistent until alternate fix to orphan data can be implemented.

[#187373461]


peterhaochen47:
The test is failing because there are still many leftovers from #711. Started a new PR: https://github.com/cloudfoundry/credhub/pull/721 that explicitly reverts the prior triggers-related commits first, and then add the commit in this current PR.

PR #721, opened 2024-04-04
fix: CredHub might fail to start due to database trigger creation permission issue

peterhaochen47:
* A prior PR (#711) is causing an issue where CredHub might fail to start if CredHub is configured with a database vendor that enforces additional access control on database triggers creation (e.g. Amazon RDS), as the prior PR's approach will attempt to create database triggers at start time. This current PR reverts the changes in the prior PR to resolve the issue while we seek for alternative solutions. 
* This PR also adds new migration files to drop the triggers in case some users have successfully added the triggers, so that the database state of all users are re-aligned.
* We have tested via local simulation that this PR's approach works for all of the below 3 scenarios:
  * Users who have not upgraded to CredHub v2.12.67 (the version of CredHub containing changes in #711)
  * Users who have successfully upgraded to CredHub v2.12.67 (all non-Amazon RDS + Amazon RDS users who somehow know to adjust access control & trigger)
  * Users who failed to upgrade to CredHub v2.12.67 due to the described issue (because they are using Amazon RDS & have not adjusted access control)


hsinn0 (approved):
Changes look good. The script file additions and deletions are as expected. I suppose that we did not have a chance to test with Amazon db, but that is OK. We can release and let TAS test it.

PR #723, opened 2024-04-08
fix: postgres script error

hsinn0:
- Fixed the drop operation order for the following error:
```
ERROR: cannot drop function del_credential_version_encrypted_value() because other objects depend on it
  Detail: trigger tr_credential_version_post_del on table credential_version depends on function del_credential_version_encrypted_value()
```


swalchemist (approved):
Interesting, I hadn't noticed that we used functions with the triggers.

PR #725, opened 2024-04-11
fix: CVE in `io.netty.netty-codec-http`

hsinn0:
- Overrode spring-boot-managed `io.netty` dependency with latest patch version.

[#187358314]


Tallicia (approved):
This looks appropriate

Issue #231, opened 2022-01-24
CredHub accumulates orphaned data in the the `encrypted_value` table


dueckminor:
What version of the credhub server you are using?
2.11.1

What version of the credhub cli you are using?
2.9.0

If you were attempting to accomplish a task, what was it you were attempting to do?
We rotate credentials regularly, and have seen that the size of the CredHub DB is growing constantly. The DB size is growing even if we rotate a credential by deleting and setting it (this deletes old credential versions). 

What did you expect to happen?
I would expect, that the credhub database doesn't grow so fast. We had databases using 15 GB storage. After I have manually deleted the orphaned entries the database size was only 40 MB.

What was the actual behaviour?
Values it the `encrypted_value` table are worthless if there are no entries in the `credential_version`, `password_credential` or `user_credential` table, because nobody can know to which credential the encrypted value belongs to.

If I generate a password credential like this:
```
credhub generate --name=/foo --type=password
```
I get two additional entries in the `encrypted_value` table. One for the value and one for the parameters. But If I delete the credential, no values will be deleted from the `encrypted_value` table.

To get rid of these entries, you may use the following SQL query (use at your own risk!):
```
DELETE FROM encrypted_value
WHERE
    NOT EXISTS (select 1 from credential_version where encrypted_value_uuid=encrypted_value.uuid)
AND
    NOT EXISTS (select 1 from password_credential where password_parameters_uuid=encrypted_value.uuid)
AND
    NOT EXISTS (select 1 from user_credential where password_parameters_uuid=encrypted_value.uuid)
```
But if the database has already a huge size, this will take some time (It took about a week to delete all orphaned entries)
By adding some indexes you could speed it up dramatically:
```
CREATE INDEX CONCURRENTLY IF NOT EXISTS credential_version_encrypted_value_uuid_idx ON credential_version USING btree (encrypted_value_uuid);
CREATE INDEX CONCURRENTLY IF NOT EXISTS password_credential_password_parameters_uuid_idx ON password_credential USING btree (password_parameters_uuid);
CREATE INDEX CONCURRENTLY IF NOT EXISTS user_credential_password_parameters_uuid_idx ON user_credential USING btree (password_parameters_uuid);
```

But that is just to get rid of existing entries. I would expect that the encrypted values get deleted together with the corresponding entries in the other tables.

Please confirm where necessary:
* [x] I have included a log output (N/A)
* [x] My log includes an error message (N/A)
* [X] I have included steps for reproduction

[...]

PR #726, opened 2024-04-13
187367323 dbs not allowing trigger creation

hsinn0:
Re-implemented fix for https://github.com/cloudfoundry/credhub/issues/231 with JPA cascade-remove instead of database triggers. See commit messages for further details.

### DIFF
--- a/components/credentials/src/main/kotlin/org/cloudfoundry/credhub/entity/Credential.kt
+++ b/components/credentials/src/main/kotlin/org/cloudfoundry/credhub/entity/Credential.kt
@@ -5,10 +5,13 @@ import org.cloudfoundry.credhub.audit.AuditableCredential
 import org.cloudfoundry.credhub.constants.UuidConstants.Companion.UUID_BYTES
 import org.hibernate.annotations.GenericGenerator
 import java.util.UUID
+import javax.persistence.CascadeType
 import javax.persistence.Column
 import javax.persistence.Entity
+import javax.persistence.FetchType
 import javax.persistence.GeneratedValue
 import javax.persistence.Id
+import javax.persistence.OneToMany
 import javax.persistence.Table
 
 @Entity
@@ -20,6 +23,11 @@ class Credential : AuditableCredential {
     @GeneratedValue(generator = "uuid2")
     @GenericGenerator(name = "uuid2", strategy = "uuid2")
     override var uuid: UUID? = null
+
+    @OneToMany(cascade = [CascadeType.REMOVE],
+        mappedBy = "credential", fetch = FetchType.EAGER)
+    var credentialVersions: MutableList<CredentialVersionData<*>> =
+        mutableListOf();
 
     @Column(nullable = false)
     override var name: String? = null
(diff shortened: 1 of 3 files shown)


components/credentials/src/main/kotlin/org/cloudfoundry/credhub/entity/Credential.kt
```diff
@@ -21,6 +24,13 @@ class Credential : AuditableCredential {
     @GenericGenerator(name = "uuid2", strategy = "uuid2")
     override var uuid: UUID? = null
 
+    @OneToMany(
+        cascade = [CascadeType.REMOVE],
+        mappedBy = "credential", fetch = FetchType.EAGER
+    )
+    var credentialVersions: MutableList<CredentialVersionData<*>> =
+        mutableListOf()
```
bruce-ricard, line 32: Does this mean that a new column will be added to the `credential` table, storing all of the credential versions associated to it?

hsinn0: No. That relationship is already in the database table schema via the FK from `credential_version` to `credential`. This JPA layer causes no change to the existing database schema. Otherwise there will be an issue for the credhub upgrade. Verified it with both fresh credhub database and upgraded credhub database.

bruce-ricard: Oh that's great. So what's the impact of this change? Will the server need to keep more data in memory? Like a `Map<Credential, List<Version>>` of some sort? Or is this again all dealt with through the DB's FKs?

It looks really magic. You just added this field, it appears to be never used anywhere, but it fixes the issue we've been having?

hsinn0: > Will the server need to keep more data in memory? Like a Map<Credential, List<Version>> of some sort?

Likely so to some degree.

> It looks really magic. 

There is no magic. Just more understanding.

> but it fixes the issue we've been having?

Right, that is the whole point of this PR.

hsinn0: Actually, there is a small magic that surprised me. I thought that there would be issues deleting `encrypted_value` associated with `user_credential` or `password_credential`, but there were all working fine, too.

components/credentials/src/test/java/org/cloudfoundry/credhub/services/DefaultCredentialVersionDataServiceTest.java
```diff
@@ -317,6 +319,7 @@ public void delete_onACredentialName_deletesAllCredentialsWithTheName() {
   }
 
   @Test
+  @Transactional(propagation = Propagation.NEVER)
```
bruce-ricard, line 322: I don't understand this. I read your commit message about it, but I still don't really get it. Does it force the test to run more serially or something?

peterhaochen47 (approved):
LGTM, tried it & it worked.

PR #731, opened 2024-04-19
Bump spring-framework to 5.3.34

peterhaochen47:


bruce-ricard (approved):
Why is dependabot not doing this?

peterhaochen47:
> Why is dependabot not doing this?

See: https://github.com/cloudfoundry/credhub/blob/7e8a7345041c1b378e5b119b26e193802656001f/build.gradle#L49
Not sure, perhaps the bot does not recognize these pinnings (which are really spring config syntax, not gradle syntax). This is a cost of not being on a supported spring boot - that we have to keep overriding it.

PR #748, opened 2024-06-07
187452409 postgres fk idx

hsinn0:
Tried creating the index with 1 million records, but did not notice significant performance issues for creating the index during Flyway migration. It was done in 1 seconds in my dev env. This does not necessarily mean that there won't be no slowdown during the index creation in the field, as my local environment is simple and there was no active database transaction while the index was created. As the implementation is supposed to create index without lock, we will just have to release it and hope that it works smoothly for most deployments our there.


Tallicia (approved):
Technically this looks good. Not clear on the issue this links back to

PR #751, opened 2024-06-18
fix: api docs build failure

peterhaochen47:
- problem: script `/scripts/generate_documentation_snippets.sh`
fails with:
```
In plugin 'org.asciidoctor.convert' type 'org.asciidoctor.gradle.AsciidoctorTask' property 'logDocuments' is annotated with @optional but that is not allowed for 'Console' properties.
```

- It looks like our asciidoctor lib is out of date (EOL) and not compatible
with the new gradle. Hence, taking advice from
https://github.com/asciidoctor/asciidoctor-gradle-plugin/issues/600,
migrate to a supported lib instead (`org.asciidoctor.jvm.convert`);
and use the latest version of that.

- The changes in this commit follows this sample app: https://github.com/spring-projects/spring-restdocs-samples/blob/c98c705ad9043aa19db96ce29160e0fbdfeaf8bc/restful-notes-spring-hateoas/build.gradle

- note: The directory changes in `credhub/backends/credhub/src/docs/asciidoc/index.adoc`
is required as part of this migration/upgrade; without these changes,
the rendered doc html will not correctly import those "snippets". The
new lib also seems to output the resulting html to a different dir,
so update tasks and readme to reflect that.

[#187494474]

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ The CredHub API can generate API documentation by running its test suite (via Sp
 ./scripts/generate_documentation_snippets.sh
 ```
 
-CredHub API documentation will be built as an html file in the CredHub backend gradle subproject build directory: `backends/credhub/build/asciidoc/html5`.
+CredHub API documentation will be built as an html file in the CredHub backend gradle subproject build directory: `backends/credhub/build/docs/asciidoc/index.html`.
 
 ### Development Configuration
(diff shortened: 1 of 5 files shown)


coolgang123:
Looks good to me. I don't have privilege to approve it.

PR #752, opened 2024-06-20
dependabot to ignore protobuf bump

iprotsiuk:
We need to stay with protobuf-java & protoc 3.25.x as the latest grpc 1.62.2 still depends on protobuf-java & protoc 3.25.x
this can be removed whenever new version of grpc that supports protobuf 4.x.x is released
https://github.com/grpc/grpc-java/issues/11015


peterhaochen47 (changes requested):
Cool. Suggestion: putting the relevant issue (https://github.com/grpc/grpc-java/issues/11015) in the comment here. So in the future people have an easy way to track if this ignore can be removed.

iprotsiuk:
added to a pr description and comment

peterhaochen47 (approved):

PR #778, opened 2024-08-20
bump: Dependency versions that we override Spring Boot's

hsinn0:
- Started overriding Tomcat version as older versions had CVEs.
- This fixes https://jira.eng.vmware.com/browse/TASCVE-12113.
- Bumped other overriding minor/patch versions to the later compatible ones.


Tallicia (approved):
bumps look ok

PR #803, opened 2024-10-23
feat: improve DB query performance

peterhaochen47:
- queries improved for MySQL and Postgres (including get-by-cred-name, find-by-name-like, delete-by-name) by performing case-insensitive search more efficiently

details:
- product requirements
    - The credential name itself is **case-sensitive** data (you store `/FooBar`, you get `/FooBar` back). But the cred search (credhub api's get-by-cred-name, get-by-cred-name-like, find-by-name endpoints) needs to be **case-insensitive** (you search `/fooBAR`, you still get `/FooBar` back).
    - CredHub has to work with MySQL, Postgres, and H2 (though H2 is only for testing).
- problem statements (current state)
    - For MySQL, the queries currently uppercase or lowercase the search: `upper(credential.name) = upper(searched_term)` (sometimes it's uppercase, sometimes lowercase; it's not consistent), and without any relevant index, a full table scan is performed (I assume because it has to freshly compute the uppercased value of the name column for every single row).
    - For Postgres, a `lower(credential.name)` index is present, but some queries still do "uppercase" (those generated by Spring JPA lib from java code), resulting in full table scan.
- solution: Add a new generated column that computes and stores the lowercased version of the credential names, and add index on it, and search by that.
    - based on performance testing with local setup (240k cred entries in DB), this improves end-to-end performance by 3 times on average. Since this measurement includes fixed costs like local credhub cli processing time, local credhub server processing time, and local network latency, the DB query performance improvement must be much higher (more than 10 times based on MySQL team's analysis).
    - the generated column syntax in postgres & mysql does not quite work in H2 (despite H2 docs saying it works: https://h2database.com/html/features.html#generated_columns). Perhaps our H2 version is too old. But a similar syntax does work (tests passed). This is fine because we don't need to optimize H2 performance since it's just for testing.

[internal tracking: https://vmw-jira.broadcom.net/browse/TPCF-26571]

### DIFF
--- a/applications/credhub-api/src/main/resources/db/migration/h2/V59__add_lowercase_credential_name_column_and_index.sql
+++ b/applications/credhub-api/src/main/resources/db/migration/h2/V59__add_lowercase_credential_name_column_and_index.sql
@@ -1,0 +1,5 @@
+ALTER TABLE credential
+    ADD COLUMN name_lowercase VARCHAR(1024) AS LOWER(name);
+
+CREATE INDEX credential_name_lowercase
+    ON credential(name_lowercase);
--- a/applications/credhub-api/src/main/resources/db/migration/mysql/V57__add_lowercase_credential_name_column.sql
+++ b/applications/credhub-api/src/main/resources/db/migration/mysql/V57__add_lowercase_credential_name_column.sql
@@ -1,0 +1,2 @@
+ALTER TABLE credential
+    ADD COLUMN name_lowercase VARCHAR(1024) GENERATED ALWAYS AS (lower(name)) STORED;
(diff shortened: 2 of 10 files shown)


applications/credhub-api/src/main/resources/db/migration/h2/V59__add_lowercase_credential_name_column_and_index.sql
duanemay, line 1: Would we have issues with name_lowercase not being populated when inserting new rows on H2?

duanemay (approved):
This should be a good performance improvement on mysql and postgres. 
Just one question on H2.

applications/credhub-api/src/main/resources/db/migration/h2/V59__add_lowercase_credential_name_column_and_index.sql
peterhaochen47: I assume that name_lowercase is indeed being populated when inserting new rows. Otherwise our tests under H2 mode wouldn't be passing, since the code is searching by name_lowercase now. E.g. [this test](https://github.com/cloudfoundry/credhub/blob/275857e2bd64301abdd869850825522990fb5cd4/backends/credhub/src/test/java/org/cloudfoundry/credhub/integration/CredentialFindTest.java#L86) inserts a new row and then search for that (the underlying query is searching by the name_lowercase column).

Issue #813, opened 2024-11-11
Database migration in credhub-release 2.12.94 fails.


jochenehret:
What version of the credhub server you are using?
We are trying to upgrade to credhub-release 2.12.94 with MySQL as database.

What version of the credhub cli you are using?
Not relevant, server-side problem.

If you were attempting to accomplish a task, what was it you were attempting to do?
Upgrade cf-deployment from v44.6.0 to v44.7.0. The credhub-release version is upgraded from 2.12.93 to 2.12.94.

What did you expect to happen?
Upgrade of credhub server succeeds.

What was the actual behavior?
Upgrade failed because of a flyway migration error. Output from credhub.log is:
```
2024-11-08T19:26:29.643Z [main] ....  INFO --- BaseDatabaseType: Database: jdbc:mariadb://sql-db.service.cf.internal:3306/credhub (MySQL 8.0)
2024-11-08T19:26:30.014Z [main] ....  WARN --- FlywayMigrationStrategyConfiguration: Validation failed: "Validate failed: Migrations have failed validation
Migration checksum mismatch for migration version 58
-> Applied to database : -1214956128
-> Resolved locally    : 1764759980. Either revert the changes to the migration, or run repair to update the schema history.
Detected resolved migration not applied to database: 57. To ignore this migration, set -ignoreIgnoredMigrations=true. To allow executing this migration, set -outOfOrder=true.
Need more flexibility with validation rules? Learn more: https://rd.gt/3AbJUZE".
2024-11-08T19:26:30.014Z [main] ....  INFO --- FlywayMigrationStrategyConfiguration: Attempting to repair...
2024-11-08T19:26:30.052Z [main] ....  INFO --- VersionPrinter: Flyway Community Edition 7.15.0 by Redgate
2024-11-08T19:26:30.109Z [main] ....  INFO --- JdbcTableSchemaHistory: Repair of failed migration in Schema History table `credhub`.`flyway_schema_history` not necessary. No failed migration detected.
2024-11-08T19:26:30.154Z [main] ....  INFO --- JdbcTableSchemaHistory: Repairing Schema History table for version 58 (Description: index lowercase credential name, Type: SQL, Checksum: 1764759980)  ...
2024-11-08T19:26:30.159Z [main] ....  INFO --- JdbcTableSchemaHistory: Repairing Schema History table for version 58 (Description: index lowercase credential name, Type: SQL, Checksum: 1764759980)  ...
[...]

PR #814, opened 2024-11-11
Roll back problematic mysql db migrations (issue #813)

peterhaochen47:
This PR, as a whole, rolls back all the DB migrations that caused* #813 but let the postgres migrations (which do not have problems) stay so as to not create unnecessary divergence in users's postgres DB states. 

*root cause analysis: 
- [Credhub version 2.12.69](https://github.com/pivotal/credhub-release/releases/tag/2.12.69) removes some problematic migrations (mysql migration v57, v58). However, some consumers have already installed prior credhub versions (e.g. credhub 2.12.68) that do contain those migrations, so their flyway history table recorded such. 
- Then #803 and #804, forgetting / not having taken that context into account, added new migrations also numbered with v57, v58; despite having different migration file names, flyway is still not happy with those repeats of migration numbers for those consumers who have installed credhub 2.12.68 for example, causing issue #813.
- Hence, remove those problematic mysql migrations for now as a temporary measure, as the removal of the mysql v57, v58 migration numbers is likely a necessity no matter what the eventual solution for re-applying #803 and #804 is.

### DIFF
--- a/applications/credhub-api/src/main/resources/db/migration/h2/V59__add_lowercase_credential_name_column_and_index.sql
+++ b/applications/credhub-api/src/main/resources/db/migration/h2/V59__add_lowercase_credential_name_column_and_index.sql
@@ -1,5 +1,0 @@
-ALTER TABLE credential
-    ADD COLUMN name_lowercase VARCHAR(1024) AS LOWER(name);
-
-CREATE INDEX credential_name_lowercase
-    ON credential(name_lowercase);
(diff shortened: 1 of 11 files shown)


hsinn0:
> forgetting / not having taken that context into account

I see. So for the change in 2.12.69, which I think that we deleted the migration sql files, it would have been better if we just changed the content of the sql file instead, i.e. to no-op. Then could see that the last-used migration script verison number in the source.

hsinn0 (commented):
It seems that this deleting migration sql files is going to work. However, can you actually test the upgrade scenario if you have not done so?

peterhaochen47:
Upgrade scenario testing results:
10c72953a4106466b1f55b18c2883e3da56a0f17

peterhaochen47:
Likely closing this in favor of https://github.com/cloudfoundry/credhub/pull/815

peterhaochen47:
closing this in favor of https://github.com/cloudfoundry/credhub/pull/815

PR #846, opened 2025-02-07
doc: Addtional comment on running CredHub with local UAA

hsinn0:


hsinn0:
> By "Alternatively", do you mean that instead of using 'docker run... ' you can instead start UAA by using the './gradlew run' as listed in UAA's README?

Yes, do you think that it is not clear enough?

joeeltgroth (approved):
lgtm

PR #849, opened 2025-02-14
New oauth2 jwt decoder for both test and run

hsinn0:
See the commit message for the change details.


joeeltgroth:
Is the strToRsaPublicKey method necessary because when key_value is null, @Value() cannot automatically convert to an RSAPublicKey?

joeeltgroth (approved):
Makes sense.  Changes look good

PR #852, opened 2025-02-18
Make UserContextFactoryTest pass

joeeltgroth:


hongchols:
Because I was not part of this change, I should be co-author of the commit. Suggest that you force-push with amend.

components/auth/src/main/kotlin/org/cloudfoundry/credhub/auth/UserContextFactory.kt
```diff
@@ -41,36 +42,11 @@ class UserContextFactory {
...
-//            scope = scopes.joinToString(",")
-//
-//            val additionalInformation = accessToken.additionalInformation
-//            userName = additionalInformation["user_name"] as String?
-//            userId = additionalInformation["user_id"] as String?
-//            issuer = additionalInformation["iss"] as String?
-//            validFrom = claimValueAsLong(additionalInformation)
-//            validUntil = accessToken.expiration.toInstant().epochSecond
-//        }
-
```
hongchols, line 73: Any idea what to do with this commented out block?

hongchols:
ktlint failure as follows:
```
% ./gradlew build -x test
<snip>
> Task :components:auth:ktlint
/Users/hsinn/prj/credhub/components/auth/src/test/kotlin/org/cloudfoundry/credhub/auth/UserContextFactoryTest.kt:13:1: Wildcard import (standard:no-wildcard-imports)
/Users/hsinn/prj/credhub/components/auth/src/test/kotlin/org/cloudfoundry/credhub/auth/UserContextFactoryTest.kt:25:1: Wildcard import (standard:no-wildcard-imports)
/Users/hsinn/prj/credhub/components/auth/src/test/kotlin/org/cloudfoundry/credhub/auth/UserContextFactoryTest.kt:157:22: A multiline expression should start on a new line (standard:multiline-expression-wrapping)
12:10:23.136 [main] WARN com.pinterest.ktlint.cli.internal.KtlintCommandLine - Lint has found errors than can be autocorrected using 'ktlint --format'

Summary error count (descending) by rule:
  standard:no-wildcard-imports: 2
  standard:multiline-expression-wrapping: 1

> Task :components:auth:ktlint FAILED
<snip>
```
As discussed previously, our current minimum requirement for `new-oahut2` branch is that build must pass.
Fix the error, and after that, make sure that credhub server starts fine and you can log in to that with `credhub-cli`.

hongchols (changes requested):

joeeltgroth:
> Because I was not part of this change, I should not be co-author of the commit. Suggest that you force-push with amend.

Have removed co-author

joeeltgroth:
> ktlint

Have fixed ktlint and have verified that server starts and can login with credhub-cli

components/auth/src/main/kotlin/org/cloudfoundry/credhub/auth/UserContextFactory.kt
joeeltgroth: I think there's two things in the commented out block: 1) the maybeToken, and 2) the fetching of additionalInformation from the accessToken. 

I think the 2nd part is taken care of by fetching the userName, userId, etc from the claims now. 

I'm uncertain about that maybeToken however.  I find only one test that exercises it with a value other than null (https://github.com/cloudfoundry/credhub/blob/0f9813d023c785044b09cffdd05b86338dff3a9f/components/auth/src/test/kotlin/org/cloudfoundry/credhub/auth/UserContextFactoryTest.kt#L65) and that test passes. Perhaps the test is insufficient?

joeeltgroth: Oh, the test was mocking ResourceServerTokenServices.readAccessToken, when "tokenValue"... need to dig further.

hsinn0:
Oops. Earlier, I mistakenly did my review with in my personal account, so I was getting notification emails in my personal email address.

components/auth/src/main/kotlin/org/cloudfoundry/credhub/auth/UserContextFactory.kt
hsinn0: OK, so I guess that I need to wait for this to come back to review again.

joeeltgroth: I concluded that the "maybeToken" wasn't necessary because the Jwt is always converted to claims, which we then convert to UserContext. I cleaned up by removing the parameter entirely in createUserContext. Tests that are the focus of this PR still passing.

components/auth/src/main/kotlin/org/cloudfoundry/credhub/auth/UserContextFactory.kt
```diff
@@ -7,70 +7,32 @@ import org.springframework.security.oauth2.server.resource.authentication.JwtAut
...
         val claims = jwt.claims
         val grantType = claims["grant_type"] as String
         val clientId = claims["client_id"] as String
         val userId = claims["user_id"] as String?
         val userName = claims["user_name"] as String?
         val issuer = claims["iss"] as String?
-        val validFrom: Long = 0
-        val validUntil: Long = 0
+        val validFrom: Long = (claims["iat"] as Instant).toEpochMilli()
+        val validUntil: Long = (claims["exp"] as Instant).toEpochMilli()
```
hsinn0, line 32: I think that both of these need to be:
```
(claims["iat"] as Instant).epochSecond
```

Both "MUST be a number containing a NumericDate value" per the [RFC](https://tools.ietf.org/html/rfc7519#section-2) mentioned in the deleted comment (line number 16). And `NumericDate` definition in the RFC is as follows: 
> NumericDate
      A JSON numeric value representing the number of seconds from
      1970-01-01T00:00:00Z UTC until the specified UTC date/time,
      ignoring leap seconds.

hsinn0 (approved):

PR #857, opened 2025-02-26
Use Spring Security 5 issuer validation

joeeltgroth:


hsinn0 (commented):
I did not realize that there were multiple test cases that expects unauthorized response with detailed messages. I thought that we were dealing with only one case. Let me have some time to review them.

hsinn0 (approved):
However, as long as we plan to revisit the response expectation changes in another story, I am OK with this change.

PR #860, opened 2025-02-27
Convert AuthWithoutOAuthConfiguration.kt

joeeltgroth:
Reasoning for deleting this file:
    - it is duplicate of AuthConfiguration with some pieces missing.
    - the default configuration of the BOSH release uses AuthConfiguration
      and not this. (security.oauth2.enabled = true)
    - there isn't evidence of security.oauth2.enabled being set to false
      anyplace in credhub code
    - So, it is probably for development purposes and can be recreated,
      if needed easily by re-duplicating AuthConfiguration
    - It contains deprecated Spring Security 2.5.x dependencies that
      need to be eliminated


hsinn0:
> - the default configuration of the BOSH release uses AuthConfiguration
>   and not this. (security.oauth2.enabled = true)
> - there isn't evidence of security.oauth2.enabled being set to false
>   anyplace in credhub code

Not sure if that is enough evidence of the configuration class not being used. I think that we need to check if it can be configured in TAS. I will try to look into it more later (maybe tomorrow).

hsinn0:
BTW, can you run unit test with this change locally if you have not done so and provide the result?

joeeltgroth:
Local test results: 

./gradlew test
BUILD SUCCESSFUL in 3m 34s
109 actionable tasks: 33 executed, 76 up-to-date

./gradlew build -x test
BUILD SUCCESSFUL in 43s
132 actionable tasks: 22 executed, 110 up-to-date

./scripts/start_server.sh
Server started and 'credhub' login for both a username and a client login successfully

hsinn0:
Observed credhub behavior with security.oauth.enabled=false.

In both credhub `main` branch and `new-oauth2` branch, the result was the same as follows:
```
% credhub login -s https://localhost:9000 --client-name=credhub_client --client-secret=secret --skip-tls-validation 
invalid_token: Invalid access token
```
So it did not function properly, but that could be because my local setup was not proper.

With `AuthWithoutOAuthConfiguration` removed in new-oatuh2 branch, we get errors as follows:
```
% credhub login -s https://localhost:9000 --client-name=credhub_client --client-secret=secret --skip-tls-validation
Error connecting to the targeted API: "The response body could not be decoded: unexpected end of JSON input". Please validate your target and retry your request.
```
So I do not think that just removing `AuthWithoutOAuthConfiguration` is a solution. Either we need to look into it more now or come back to this after other stories and maybe after gaining more knowledge on how mtls works in credhub.

See the related Jira story for additional info including credhub server logs.

joeeltgroth:
True, the bosh release does allow 'credhub.authentication.uaa.enabled' to be set to false, so we should convert this class instead of removing it.  Have provided a converted version, but it will need updating after we finish work on error handling.

hsinn0 (approved):
Approving with the following conditions:
- We keep the related story in progress until all the questions are answered.
- The one WIP commit in cleared, or we merge this as a single commit to new-aout2 branch with a new commit message.

hsinn0 (approved):

PR #868, opened 2025-03-13
feat: Provide OAuth error info in response body.

hsinn0:
- As we did in the existing credhub implemenmtation.
- However, instead of providing customized `error_description` string as we did in the existing credhub implementation, provide the `error_description` from `spring-security`, i.e. the same ones that are provided in the response header.
- It is because, unlikely the existing credhub implementation, we cannot tell exact kind of error based on the cause exceptions. For example, invalid issuer and expired token have the same cause exception but with specific error messages.

Making it as Draft PR for now, as I want to perform additional tasks as follows:
- Test with credhub-cli.
- Take another look at the option to provide customized `error_description` string as we did in the existing credhub implementation. I did spent quite some time on this already but ...

### DIFF
--- a/components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
+++ b/components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
@@ -1,0 +1,46 @@
+package org.cloudfoundry.credhub.config.auth;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
+import org.springframework.security.oauth2.core.OAuth2Error;
+import org.springframework.security.oauth2.server.resource.web.BearerTokenAuthenticationEntryPoint;
+import org.springframework.security.web.AuthenticationEntryPoint;
+
+import com.nimbusds.jose.shaded.gson.JsonObject;
+
+/**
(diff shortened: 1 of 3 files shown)


components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
```diff
@@ -0,0 +1,46 @@
...
+public class OAuth2AuthenticationExceptionHandler
+        implements AuthenticationEntryPoint {
+
+    @Override
+    public void commence(HttpServletRequest request,
+                         HttpServletResponse response,
+                         AuthenticationException authException)
+            throws IOException {
+        BearerTokenAuthenticationEntryPoint btaep =
+                new BearerTokenAuthenticationEntryPoint();
```
joeeltgroth, line 30: Would it work to instantiate once as a bean and re-use rather than creating a new instance each commence()?

joeeltgroth (changes requested):

components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
hsinn0: Oops. That was oversight. Thank you for catching that.

hsinn0: Fixed it with the new commit. Also marking this PR as ready to merge as you suggested in Gspace. Will do the additional tasks mentioned above as part of the associated story.

joeeltgroth: lgtm

joeeltgroth (approved):

Issue #872, opened 2025-03-14
Bug: Clean Project Startup


jbuecher:
**What version of the credhub server you are using?**
main Branch

**What version of the credhub cli you are using?**
2.9.43

**If you were attempting to accomplish a task, what was it you were attempting to do?**
After cloning a fresh copy from the main branch of the repository, I'd like to run the `scripts/start_server.sh` to compile and start the application.

**What did you expect to happen?**
As before the bump to Gradle version 8.13, I expected that the service starts after running the script.

**What was the actual behavior?**
After compilation of the Kotlin and Java classes, the service failed to start with the following dump:

<details>
<summary>Log Dump</summary>

```text
> Task :applications:credhub-api:bootRun FAILED

2025-03-14T12:54:03.848Z [background-preinit] INFO  org.hibernate.validator.internal.util.Version - HV000001: Hibernate Validator 6.2.5.Final
2025-03-14T12:54:03.872Z [main] INFO  org.cloudfoundry.credhub.CredHubApp - Starting CredHubApp using Java 17.0.14 on HJ7V9PDWMX with PID 30324 (/Users/I552343/go/src/github.com/cloudfoundry/my-credhub-fork/applications/credhub-api/build/classes/java/main started by I552343 in /Users/I552343/go/src/github.com/cloudfoundry/my-credhub-fork)
2025-03-14T12:54:03.873Z [main] INFO  org.cloudfoundry.credhub.CredHubApp - The following 3 profiles are active: "dev", "dev-remote-uaa", "dev-h2"
2025-03-14T12:54:04.440Z [main] INFO  org.springframework.data.repository.config.RepositoryConfigurationDelegate - Bootstrapping Spring Data JPA repositories in DEFAULT mode.
2025-03-14T12:54:04.504Z [main] INFO  org.springframework.data.repository.config.RepositoryConfigurationDelegate - Finished Spring Data repository scanning in 59 ms. Found 5 JPA repository interfaces.
2025-03-14T12:54:04.793Z [main] INFO  org.springframework.boot.web.embedded.tomcat.TomcatWebServer - Tomcat initialized with port(s): 9000 (https)
2025-03-14T12:54:04.800Z [main] INFO  org.apache.coyote.http11.Http11NioProtocol - Initializing ProtocolHandler ["https-jsse-nio-9000"]
2025-03-14T12:54:04.801Z [main] INFO  org.apache.catalina.core.StandardService - Starting service [Tomcat]
[...]

PR #873, opened 2025-03-14
build(deps): Bump Java version from 17 to 21 (LTS)

jbuecher:
Due to issue described in: [issue](https://github.com/cloudfoundry/credhub/issues/872)


hsinn0:
As mentioned in https://github.com/cloudfoundry/credhub/issues/872, we will resolve the issue by bump Tomcat 9 version instead of Java version. For Java version bump, we have an internal task planned, which includes more than just credhub source update.

PR #874, opened 2025-03-14
Add exception handling for bad MTLS certificate

joeeltgroth:


components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
```diff
@@ -6,6 +6,7 @@
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.springframework.security.authentication.InsufficientAuthenticationException;
```
hsinn0, line 9: This is not used.

components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
```diff
@@ -30,17 +31,29 @@ public void commence(HttpServletRequest request,
...
+        }
+
+        var jsonObject = new JsonObject();
+
         if (authException instanceof OAuth2AuthenticationException) {
-            OAuth2Error error = ((OAuth2AuthenticationException) authException)
-                    .getError();
-            response.setContentType("application/json");
-            var jsonObject = new JsonObject();
+
```
hsinn0, line 41: Why do we have extra empty line here?

components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
```diff
@@ -30,17 +31,29 @@ public void commence(HttpServletRequest request,
...
+            OAuth2Error error = ((OAuth2AuthenticationException) authException).getError();
             jsonObject.addProperty("error", error.getErrorCode());
             jsonObject.addProperty("error_description", error.getDescription());
-            try (PrintWriter writer = response.getWriter()) {
-                writer.write(jsonObject.toString());
-                writer.flush();
-            }
+
```
hsinn0, line 45: Extra empty line again here and 3 more below.

components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
```diff
@@ -30,17 +31,29 @@ public void commence(HttpServletRequest request,
...
+        } else {
```
hsinn0, line 46: So what are the types of `authException`s that are not `new-oauth2-mtls-exception`?

hsinn0 (changes requested):

components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
joeeltgroth: The one that the 2 mtls acceptance tests were producing is InsufficientAuthenticationException.

components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
hsinn0: Unused import also causes `checkStyle` failure. I suggest doing `build -x test` locally at least to catch this kind of oversight.

components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
hsinn0: Are we sure that we want to populate the response body not just for `InsufficientAuthenticationException` but for all `AuthenticationException`?

components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
```diff
@@ -30,17 +30,24 @@ public void commence(HttpServletRequest request,
...
+            jsonObject.addProperty("error_description", authException.getMessage());
         }
+
+        response.setContentType("application/json");
+        try (PrintWriter writer = response.getWriter()) {
+            writer.write(jsonObject.toString());
+            writer.flush();
+        }
+
```
hsinn0, line 51: Another extra empty line here.

hsinn0 (changes requested):
BTW, after resolving all review comments, please re-request review so the reviewer can know that it is ready for review again.

components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
joeeltgroth: Unsure, figured it was a more consistent response content behavior to reply with the Exception message for all.  I don't currently see a downside.

components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
```diff
@@ -29,18 +29,21 @@ public void commence(HttpServletRequest request,
                          AuthenticationException authException)
             throws IOException {
         btaep.commence(request, response, authException);
-
+        if (authException == null) {
+            return;
+        }
```
hsinn0, line 34: I missed this change earlier. We do not need null check here.

joeeltgroth: Removed

hsinn0 (approved):

PR #876, opened 2025-03-18
Fix the InsufficientAuthenticationException handling

joeeltgroth:
Make the InsufficientAuthenticationException handling properly handle the mtls_test.go acceptance test expectations.

### DIFF
--- a/components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
+++ b/components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
@@ -36,6 +36,7 @@ public class OAuth2AuthenticationExceptionHandler
             jsonObject.addProperty("error", error.getErrorCode());
             jsonObject.addProperty("error_description", error.getDescription());
         } else if (authException instanceof InsufficientAuthenticationException) {
+            jsonObject.addProperty("error", "invalid_token");
             jsonObject.addProperty("error_description", authException.getMessage());
         }
         response.setContentType("application/json");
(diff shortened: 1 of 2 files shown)


components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
```diff
@@ -36,6 +36,7 @@ public void commence(HttpServletRequest request,
             jsonObject.addProperty("error", error.getErrorCode());
             jsonObject.addProperty("error_description", error.getDescription());
         } else if (authException instanceof InsufficientAuthenticationException) {
+            jsonObject.addProperty("error", "invalid_token");
```
hsinn0, line 39: So, this change is to satisfy the expectation of the test cases in `credhub_acceptance_tests`, correct?
I do not think  `invalid_token` is proper value for `InsufficientAuthenticationException`, though. How about we set it to "access_denied" (picked from https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1) and change the test expectations instead?

hsinn0 (changes requested):

components/auth/src/main/java/org/cloudfoundry/credhub/config/auth/OAuth2AuthenticationExceptionHandler.java
joeeltgroth: Yes, this change was attempting to match the credhub_acceptance_tests exactly.  I have now added a commit here that uses the standard OAuth2 Error Response.

hsinn0 (approved):

PR #912, opened 2025-05-15
Various dependency refactors and organization

peterhaochen47:
- See individual commits

# notes
- out of scope: the deps brought in by spring boot by default that are major version(s) behind what we explicitly pinned / bring in (don't wanna downgrade anything by a major version, e.g. flyway, even though it's probably not a big deal, since spring boot often values stability).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,18 +8,8 @@ updates:
   target-branch: main
   open-pull-requests-limit: 100
   ignore:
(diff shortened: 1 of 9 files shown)


hsinn0 (commented):
`build.gradle` changes look fine. Are you planning to make change in `.github/dependabot.yml` in separate commit?

peterhaochen47:
@hsinn0 Oh yeah, I missed that part of the ACs. I just pushed a new commit.

hsinn0 (approved):

PR #915, opened 2025-05-16
Rewrote a test case.

hsinn0:
- As the existing code that mock request body object is not working after spring boot3 migration.
- Modified the requst content to actually cause the exception to be thrown in the controller.

### DIFF
--- a/backends/credhub/src/test/java/org/cloudfoundry/credhub/integration/v1/credentials/CredentialsTypeSpecificSetIntegrationTest.java
+++ b/backends/credhub/src/test/java/org/cloudfoundry/credhub/integration/v1/credentials/CredentialsTypeSpecificSetIntegrationTest.java
@@ -1,7 +1,6 @@
 package org.cloudfoundry.credhub.integration.v1.credentials;
 
 import java.io.IOException;
-import java.io.InputStream;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


backends/credhub/src/test/java/org/cloudfoundry/credhub/integration/v1/credentials/CredentialsTypeSpecificSetIntegrationTest.java
```diff
@@ -499,13 +494,9 @@ public void validationExceptionsAreReturnedAsErrorMessages() throws Exception {
       .content("{" +
         "\"name\":\"" + CREDENTIAL_NAME + "\"," +
         "\"type\":\"" + parametizer.credentialType + "\"," +
-        "\"value\":" + parametizer.credentialValue +
+        "\"value\":" +  // Empty vaiue here will cause ParameterizedValidationException(ErrorMessages.BAD_REQUEST) to be thrown.
```
peterhaochen47, line 497: typo in "value", otherwise LGTM

peterhaochen47 (approved):

backends/credhub/src/test/java/org/cloudfoundry/credhub/integration/v1/credentials/CredentialsTypeSpecificSetIntegrationTest.java
hsinn0: Fixed with force push.

PR #918, opened 2025-05-20
remove: Use of deprecated `WebMvcConfigurer.setUseTrailingSlashMatch()`

hsinn0:
- Changed to use `UrlHandlerFilter` instead.
- To keep accepting URLs with trailing slashes.


peterhaochen47 (approved):
Approved assuming that this works (can be / will be verified in credhub acceptance tests)

hsinn0:
> Approved assuming that this works (can be / will be verified in credhub acceptance tests)

Yeah, I did some manual verification, too, i.e. calling endpoints with and without backslashes.

PR #928, opened 2025-05-28
upgrade: Replace use of deprecated `X509Certificate.subjectDN`

hsinn0:
- With `X509Certificate.subjectX500Principal`.


hsinn0:
This is for TNZ-44060.

acosta11 (approved):
LGTM, checked https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/security/cert/X509Certificate.html and confirmed.

> getSubjectDN()
Deprecated.
Use getSubjectX500Principal() instead.

hsinn0:
2 unit test case was failing because of the test mock mismatch with the changed test target. Corrected in the 2nd commit.

PR #929, opened 2025-05-28
upgrade: Replace use of deprecated `JdbcTemplate.query(String, @Nulla…

hsinn0:
…ble Object[], RowMapper<T>)`

- With `JdbcTemplate.query(String, RowMapper<T>, @Nullable Object...)`.

For TNZ-44061.


acosta11 (approved):
LGTM checked to confirm https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/jdbc/core/JdbcTemplate.html . Looks like the update is consistent with the previous method.

> \<T\> [List]\<T\> query(String sql, Object[] args, RowMapper<T> rowMapper) Deprecated.

>\<T\> List\<T\> query(String sql, RowMapper<T> rowMapper, Object... args) 
Query given SQL to create a prepared statement from SQL and a list of arguments to bind to the query, mapping each row to a result object via a RowMapper.

dlinsley (approved):
Looks good to me as well, thanks!

PR #930, opened 2025-05-29
fix: Incorrect parameter value

hsinn0:
- For `jdbcTemplate.query()`.
- Earlier commit to replace the deprecated method version had a mistake.
- Needs to pass `workerDatabaseName` to the new version of the method, not array of `workerDatabaseName`.
- Also refactored the conditional logic to remove double negation.


hsinn0:
This is to fix the mistake made in https://github.com/cloudfoundry/credhub/pull/929.

hsinn0:
> LGTM.
> 
> Makes sense that it might not automatically handle destructuring an array for you into the variadic parameter list required for the new method signature.

Actually, it was causing tests that use MySQL and PostgreSQL to fail because the condition was not working. I did not realize that until I saw CI job failures.

PR #1083, opened 2026-04-06
fix: PostgreSQL unit test hang

hsinn0:
- Added FlywayPostgresTestStripConcurrentlyConfiguration to transform SQL scripts and remove CONCURRENTLY keyword from index DDL to prevent blocking during parallel test execution.
- Updated ParallelPostgresTestDataSourceConfiguration to use HikariCP with connection pool limits so as to avoid too many connection errors. Also set the timeouts to avoid indefinite waits.
- Also updated ParallelPostgresTestDataSourceConfiguration to use non-pooled database connection for 'create database'.
- AI-Assisted.

### DIFF
--- a/components/test-support/src/test/kotlin/org/cloudfoundry/credhub/config/FlywayPostgresTestStripConcurrentlyConfiguration.kt
+++ b/components/test-support/src/test/kotlin/org/cloudfoundry/credhub/config/FlywayPostgresTestStripConcurrentlyConfiguration.kt
@@ -1,0 +1,83 @@
+package org.cloudfoundry.credhub.config
+
+import org.flywaydb.core.api.ResourceProvider
+import org.flywaydb.core.api.configuration.FluentConfiguration
+import org.flywaydb.core.api.migration.JavaMigration
+import org.flywaydb.core.api.resource.LoadableResource
+import org.flywaydb.core.internal.scanner.Scanner
+import org.springframework.boot.autoconfigure.flyway.FlywayConfigurationCustomizer
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.context.annotation.Profile
+import org.springframework.core.Ordered
+import org.springframework.core.annotation.Order
+import java.io.Reader
+import java.io.StringReader
+
+/**
+ * For unit tests against PostgreSQL, strip `CONCURRENTLY` from index DDL in Flyway SQL migrations so
+ * migrations do not block on other connections (e.g. Flyway bookkeeping). Production migrations are
+ * unchanged on disk; only the in-memory script content is transformed.
+ *
+ * Every `*.sql` resource is wrapped (Flyway relative paths omit `db/migration/postgres/`); replacements are
+ * no-ops for scripts that do not use concurrent index DDL.
+ */
+@Configuration
+@Profile("unit-test-postgres")
+class FlywayPostgresTestStripConcurrentlyConfiguration {
+    @Bean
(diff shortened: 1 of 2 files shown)


components/test-support/src/test/kotlin/org/cloudfoundry/credhub/config/FlywayPostgresTestStripConcurrentlyConfiguration.kt
```diff
@@ -0,0 +1,83 @@
...
+import java.io.Reader
+import java.io.StringReader
+
+/**
+ * For unit tests against PostgreSQL, strip `CONCURRENTLY` from index DDL in Flyway SQL migrations so
+ * migrations do not block on other connections (e.g. Flyway bookkeeping). Production migrations are
+ * unchanged on disk; only the in-memory script content is transformed.
+ *
+ * Every `*.sql` resource is wrapped (Flyway relative paths omit `db/migration/postgres/`); replacements are
+ * no-ops for scripts that do not use concurrent index DDL.
```
dlinsley, line 23: question:  curious why this needed now, is it due to changes in Flyway or recent changes to the tests?

Is this related to https://github.com/flyway/flyway/issues/3508 or https://github.com/flyway/flyway/issues/3682
workaround suggests applying config `flyway.postgresql.transactional.lock=false`

dlinsley: I see we attempt this in: https://github.com/cloudfoundry/credhub/blob/main/applications/credhub-api/src/main/java/org/cloudfoundry/credhub/config/FlywayMigrationStrategyConfiguration.java

hsinn0: > question: curious why this needed now, is it due to changes in Flyway or recent changes to the tests?

The unit-tests-postgres-* jobs in Coucourse have been failing frequently with timeouts in the past, too. It seems that the timeout is just more frequent (or consistent) these days. I've never bothered looked into those CI timeouts, thinking they are flakes, as rerunning usually made the jobs pass.
For the matter of local run in my dev box, I do not regularly run the tests against postgresql database unless there is something postgresql-specific that need to be looked into. So no idea when the problem started to occur there. It is possible that this was caused by some changes in other dependencies such as hikari pool or hibernate.
 
> Is this related to [flyway/flyway#3508](https://github.com/flyway/flyway/issues/3508) or [flyway/flyway#3682](https://github.com/flyway/flyway/issues/3682) workaround suggests applying config `flyway.postgresql.transactional.lock=false`

Not sure if our current issue is related to those old flyway issues. Sounds similar, though. And yes, we are already setting transitionalLock to false (in FlywayMigrationStrategyTestConfig for tests).

As the goal is to prevent the hang for the tests and the issue is not to be expected to occur in production, and the change only affect test run with postgresql database, I think that just removing concurrently from the create/drop index statements for test only is the way to go.

peterhaochen47: Yeah, the latest issue: https://github.com/flyway/flyway/issues/3961
Seems like there's no other solutions yet.

dlinsley (approved):
I was able to locally observe the PostgreSQL unit test hang locally.  These changes do resolve the observed hang during tests allowing them to complete locally.  Thanks for addressing the questions!

peterhaochen47 (approved):

PR #1092, opened 2026-04-15
Fix: migration hang by disabling Flyway transactional locks early

peterhaochen47:
- With Flyway 12.x, transactional advisory locks are enabled by default for PostgreSQL. These locks are taken during the validation and repair phases and are held throughout the database session. This causes non-transactional migrations using 'CREATE INDEX CONCURRENTLY' (such as V59 and V61) to hang indefinitely, as PostgreSQL does not allow concurrent index creation while other locks are held in the same session.
- The previous programmatic override in the 'FlywayMigrationStrategy' was running too late (after the validation and repair steps had already established a lock). This commit moves the logic to a 'FlywayConfigurationCustomizer', ensuring the transactional lock is disabled before Flyway performs any database operations.
- Changes applied to both the main application and test configurations to ensure consistency across production and CI environments.

ai-assisted=yes
TNZ-93146

### DIFF
--- a/applications/credhub-api/src/main/java/org/cloudfoundry/credhub/config/FlywayMigrationStrategyConfiguration.java
+++ b/applications/credhub-api/src/main/java/org/cloudfoundry/credhub/config/FlywayMigrationStrategyConfiguration.java
@@ -7,6 +7,7 @@ import org.flywaydb.core.Flyway;
 import org.flywaydb.core.api.FlywayException;
 import org.flywaydb.database.postgresql.PostgreSQLConfigurationExtension;
 import org.jetbrains.annotations.NotNull;
+import org.springframework.boot.autoconfigure.flyway.FlywayConfigurationCustomizer;
 import org.springframework.boot.autoconfigure.flyway.FlywayMigrationStrategy;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -20,21 +21,25 @@ public class FlywayMigrationStrategyConfiguration {
     private static final Logger LOGGER = LogManager.getLogger(FlywayMigrationStrategyConfiguration.class.getName());
 
     @Bean
-    public FlywayMigrationStrategy repairBeforeMigration() {
-        return flyway -> {
-            renameMigrationTableIfNeeded(flyway);
-            repairIfNecessary(flyway);
-
-            String url = flyway.getConfiguration().getUrl();
+    public FlywayConfigurationCustomizer postgresFlywayCustomizer() {
+        return configuration -> {
+            String url = configuration.getUrl();
             if (url != null && url.contains("postgresql")) {
                 // For CREATE INDEX CONCURRENTLY. See
                 // https://documentation.red-gate.com/fd/flyway-postgresql-transactional-lock-setting-277579114.html.
-                flyway.getConfigurationExtension(
+                configuration.getConfigurationExtension(
                         PostgreSQLConfigurationExtension.class).setTransactionalLock(
                         false);
                 LOGGER.trace("Set flyway.postgresql.transactional.lock to false.");
             }
+        };
+    }
 
+    @Bean
+    public FlywayMigrationStrategy repairBeforeMigration() {
+        return flyway -> {
+            renameMigrationTableIfNeeded(flyway);
+            repairIfNecessary(flyway);
             runMigration(flyway);
         };
     }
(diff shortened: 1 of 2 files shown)


dlinsley:
have we tried this approach while reverting the prior attempts: https://github.com/cloudfoundry/credhub/pull/1083

peterhaochen47:
> have we tried this approach while reverting the prior attempts: #1083

* The prior attempt was about fixing the hang in unit tests (looks like by removing the `CONCURRENTLY` in migrations), which worked. The code change there has no impact on how migrations behave in production.
* but the "production tests" (actually deploying credhub and real DBs) are still failing (see [CI](https://bosh.ci.cloudfoundry.org/teams/uaa/pipelines/credhub-edge)). And this PR attempts to address the production hangs only (I don't know if it works yet; it's kinda complicated to try so I was planning to take a short cut and just let CI verify it). 

This PR also modifies the unit tests migration strategy (aka what #1083 touched) for consistency, which I think is fair. But ideally, we don't have a separate flyway migration strategy just for unit tests, compared to production tests (#1083 introduced this separation). We should think about how to unify them if possible. For example, perhaps the approach in this PR would also address the unit tests hangs, and that would prove #1083 unnecessary.

But to answer your question, reverting the prior attempt #1083 would have no impact on the problem I'm trying to address (production hang). Though later, if the approach in this PR works, we should take a look at whether #1083 is necessary to address unit test hangs.

peterhaochen47:
# more detailed explanation

To verify that this approach works, we can look at the architectural shift in Flyway (specifically from version 9.x onwards) and how Spring Boot manages its lifecycle.

### Why the previous approach stopped working

The previous approach failed for two primary reasons related to how Flyway 12.4.0 handles configuration and database sessions:

#### 1. Configuration Freezing (Immutability)
In modern Flyway (9.x+), the `Flyway` object is created by calling `configuration.load()`. Once this object is instantiated, its internal configuration is intended to be **immutable**.
In the old code:
```java
// Inside FlywayMigrationStrategy
flyway.getConfigurationExtension(PostgreSQLConfigurationExtension.class)
      .setTransactionalLock(false);
```
While this code *appears* to modify the configuration, the internal engine components (like the `PostgreSQLAdvisoryLockTemplate`) have often already resolved and cached the configuration settings during the initial setup of the `Flyway` object. Modifying the extension after the object is live frequently fails to propagate to the low-level database session manager.

#### 2. Session Pollution (The "Repair" Trap)
In your logs, `repairIfNecessary(flyway)` was called **before** the programmatic override. 
*   **Log Line 175**: `Attempting to repair...` calls `flyway.repair()`.
*   At this moment, Flyway 12.4.0 uses its **default** configuration (`transactionalLock = true`).
*   It establishes a database connection and executes `SELECT pg_advisory_lock(...)`.
*   This lock is **session-level**. Even if you successfully changed the flag to `false` afterwards, that specific database session **already holds the lock**.
*   When `CREATE INDEX CONCURRENTLY` (Version 59/61) runs, PostgreSQL checks if the current session holds any advisory locks or transactions. Since it does, the operation hangs indefinitely waiting for the "blocking" lock to be released—which will never happen because Flyway is waiting for the index to finish.

### How the new approach (`FlywayConfigurationCustomizer`) addresses the root cause

The `FlywayConfigurationCustomizer` is a Spring Boot callback that hooks into the **construction phase** of the Flyway bean. 

#### 1. Pre-Initialization Configuration
The lifecycle now looks like this:
1.  Spring Boot creates a `FluentConfiguration` object.
2.  **`FlywayConfigurationCustomizer` is called**: We set `transactionalLock = false` on the `FluentConfiguration`.
3.  Spring Boot calls `configuration.load()` to create the `Flyway` bean.
4.  The `Flyway` bean is now born with the correct setting **baked into its core**.

#### 2. Clean Sessions from the Start
When the `FlywayMigrationStrategy` eventually runs:
*   It calls `repairIfNecessary(flyway)`.
*   Flyway checks its configuration, sees `transactionalLock = false`, and **skips the advisory lock call entirely**.
*   The database session remains "clean" (no locks, no active transactions).
*   When it reaches Version 59/61, the `CREATE INDEX CONCURRENTLY` command can proceed immediately because there are no session-level conflicts.

### Verification in Flyway Source Code
If you examine Flyway's `org.flywaydb.database.postgresql.PostgreSQLAdvisoryLockTemplate`, you will see logic similar to this:
```java
public void execute(Context context) {
    if (configuration.getPluginRegister().getPlugin(PostgreSQLConfigurationExtension.class).isTransactionalLock()) {
        // This is what caused the hang
        jdbcTemplate.execute("SELECT pg_advisory_lock(...)"); 
    }
    // ... perform operation ...
}
```
By using the `Customizer`, we ensure that the very first time this code is reached (even during a `repair` or `validate` call), the `isTransactionalLock()` check returns `false`.

**Summary:** The new approach works because it ensures the configuration is correct **before** the database session is ever opened, preventing the "polluted session" state that caused the previous hangs.

dlinsley (approved):
Thanks for the clarifications!

PR #1094, opened 2026-04-17
fix: postgres hangs in bosh director context

peterhaochen47:
- Robust Customizer: removed the getUrl() check in FlywayMigrationStrategyConfiguration.java (and the test config). Since the PostgreSQLConfigurationExtension only affects PostgreSQL anyway, we can safely set it without checking the URL.

- problem: Although we moved the logic to a FlywayConfigurationCustomizer (https://github.com/cloudfoundry/credhub/commit/9145c164c26808447357e90cf1cc0297a26b15d8), the implementation had a subtle flaw that caused it to be skipped in the BOSH deployment. In Spring Boot, when a DataSource is provided (which is the case in CredHub within the bosh director context), the FluentConfiguration.getUrl() often returns null because the connection details are encapsulated within the DataSource object rather than being set as a direct JDBC URL string on the Flyway configuration. Consequently, the if block was never executed, and the transactional lock remained enabled (default true).

ai-assisted=yes
TNZ-93146

### DIFF
--- a/applications/credhub-api/src/main/java/org/cloudfoundry/credhub/config/FlywayMigrationStrategyConfiguration.java
+++ b/applications/credhub-api/src/main/java/org/cloudfoundry/credhub/config/FlywayMigrationStrategyConfiguration.java
@@ -23,15 +23,12 @@ public class FlywayMigrationStrategyConfiguration {
     @Bean
     public FlywayConfigurationCustomizer postgresFlywayCustomizer() {
         return configuration -> {
-            String url = configuration.getUrl();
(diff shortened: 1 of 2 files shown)


peterhaochen47:
The last PR https://github.com/cloudfoundry/credhub/pull/1092 fixed most of the CI failures, but somehow when credhub is deployed within the bosh director, the test is still failing (postgres hang). This PR attempts to address that.

dlinsley (approved):
Thanks!  This makes sense to me. The configuration is Postgres specific and such shouldn't be conditional on the contents of the configuration URL.

PR #1098, opened 2026-04-20
Revert "fix: PostgreSQL unit test hang"

peterhaochen47:
This reverts commit fc37bbb029d93cad206bf29b80e82b94e495566b.

The underlying cause of the hang has been
fixed by:
- 9145c164c26808447357e90cf1cc0297a26b15d8
- b0587fd78d786724dec16fff8a86b806dd5c8e84 whose concerns are about migrations in prod
but also fix the unit test hang.

So having a bespoke config for how flyway
migration is run in unit test context is
not required anymore.


peterhaochen47:
# validation
- check out the commit **before** fc37bbb029d93cad206bf29b80e82b94e495566b
- confirm that the postgres unit tests is indeed hanging when run locally
- check out the latest main (without this PR)
- confirm that the postgres unit tests is not hanging when run locally
- checkout this PR
- confirm that the postgres unit tests is still not hanging when run locally

dlinsley (approved):
Thanks for following through with this and confirming it was no longer needed!

PR #1099, opened 2026-04-22
fix: resolve Flyway 12 nknown migration type failure 

peterhaochen47:
- ...by patching legacy SPRING_JDBC migration types
- This commit fixes a FATAL startup error encountered in CredHub 2.15.4: "Unknown migration type 'SPRING_JDBC' found in schema history." This issue prevents the application from starting in environments originally deployed before 2020, where the database history table still contains legacy migration types.

### The Problem
The 'SPRING_JDBC' migration type was deprecated in Flyway 6 and officially removed in Flyway 8. While subsequent versions (v9–v11) tolerated these orphaned records, Flyway 12 (introduced in CredHub 2.15.0) finalized the removal of the internal mapping, making the type unrecognized.

The issue became a fatal crash in the 2.15.4 release due to a configuration refactor that disabled the PostgreSQL 'transactional.lock' globally and early via a FlywayConfigurationCustomizer. In PostgreSQL, disabling this lock forces Flyway into a "minimalist" operational mode designed for non-transactional DDL. In this mode, Flyway’s metadata parser is significantly stricter and lacks the legacy fallback handlers required to process or "repair" unrecognized types. Consequently, Flyway crashes during its initial table scan before the programmatic repair() command can even be invoked.

### The Solution
To resolve this without requiring manual database intervention from users, this change implements an automated "pre-flight" database patch:
- Extends the `DatabaseLayer` and `DatatabaseLayerImpl` to include `updateSpringJdbcMigrationTypes()`.
- Executes a direct SQL update: "UPDATE flyway_schema_history SET type = 'JDBC' WHERE type = 'SPRING_JDBC'".
- Orchestrates this cleanup within `FlywayMigrationStrategyConfiguration` to ensure the records are converted before Flyway’s strict parser initializes.

This approach restores compatibility for long-lived production databases while maintaining the early-disable lock configuration required to prevent PostgreSQL migration hangs.

https://github.com/pivotal/credhub-release/issues/427

### DIFF
--- a/applications/credhub-api/src/main/java/org/cloudfoundry/credhub/config/DatabaseLayer.java
+++ b/applications/credhub-api/src/main/java/org/cloudfoundry/credhub/config/DatabaseLayer.java
@@ -6,4 +6,5 @@ public interface DatabaseLayer {
     boolean oldFlywayMigrationTableExists() throws SQLException;
     boolean newFlywayMigrationTableExists() throws SQLException;
     void updateOldMigrationTableName() throws SQLException;
+    void updateSpringJdbcMigrationTypes() throws SQLException;
 }
--- a/applications/credhub-api/src/main/java/org/cloudfoundry/credhub/config/DatatabaseLayerImpl.java
+++ b/applications/credhub-api/src/main/java/org/cloudfoundry/credhub/config/DatatabaseLayerImpl.java
@@ -49,4 +49,19 @@ public class DatatabaseLayerImpl implements DatabaseLayer {
             );
         }
     }
+
+    @Override
+    public void updateSpringJdbcMigrationTypes() throws SQLException {
+        if (newFlywayMigrationTableExists()) {
+            LOGGER.info("Updating SPRING_JDBC migration types to JDBC in '{}'", NEW_HISTORY_TABLE_NAME);
+            try (Statement stmt = connection.createStatement()) {
+                stmt.execute(
+                        String.format(
+                                "UPDATE %s SET type = 'JDBC' WHERE type = 'SPRING_JDBC'",
+                                NEW_HISTORY_TABLE_NAME
+                        )
+                );
+            }
+        }
+    }
 }
--- a/applications/credhub-api/src/main/java/org/cloudfoundry/credhub/config/FlywayMigrationStrategyConfiguration.java
+++ b/applications/credhub-api/src/main/java/org/cloudfoundry/credhub/config/FlywayMigrationStrategyConfiguration.java
@@ -36,6 +36,7 @@ public class FlywayMigrationStrategyConfiguration {
     public FlywayMigrationStrategy repairBeforeMigration() {
         return flyway -> {
             renameMigrationTableIfNeeded(flyway);
+            updateSpringJdbcMigrationTypes(flyway);
             repairIfNecessary(flyway);
             runMigration(flyway);
         };
@@ -52,6 +53,17 @@ public class FlywayMigrationStrategyConfiguration {
         } catch (SQLException ex) {
             LOGGER.fatal("Error renaming migration table.");
             throw new FlywayException("Error renaming migration table", ex);
+        }
+    }
+
+    @VisibleForTesting
+    void updateSpringJdbcMigrationTypes(@NotNull Flyway flyway) {
+        try (Connection connection = getConnection(flyway)) {
+            DatabaseLayer databaseLayer = new DatatabaseLayerImpl(connection);
+            databaseLayer.updateSpringJdbcMigrationTypes();
+        } catch (SQLException ex) {
+            LOGGER.fatal("Error updating SPRING_JDBC migration types.");
+            throw new FlywayException("Error updating SPRING_JDBC migration types", ex);
         }
     }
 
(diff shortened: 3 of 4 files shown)


peterhaochen47:
# detailed analysis

The failure in **CredHub 2.15.4** is the culmination of a long-standing technical debt regarding Flyway migration types and a recent architectural change that inadvertently stripped away a years-old "safety net."

### 1. The History of `SPRING_JDBC`: A Long-Abandoned Standard
The `SPRING_JDBC` migration type has been "dead weight" in the Flyway ecosystem for over half a decade:
*   **The Origin**: It was used for migrations implementing the `SpringJdbcMigration` interface, allowing migrations to use a Spring-managed `JdbcTemplate`.
*   **The Deprecation (2019)**: Flyway 6.0 deprecated the interface in favor of `BaseJavaMigration` (which uses the standard `JDBC` type).
*   **The Removal (2021)**: **Flyway 8.0** officially removed the `SPRING_JDBC` type handler from its public API and internal enums.
*   **CredHub’s "Luck"**: CredHub converted its migrations to Kotlin/`BaseJavaMigration` around 2020. However, databases deployed before that date still had the string `'SPRING_JDBC'` written into their history tables. For years (Flyway 8 through 11), Flyway maintained internal "fuzzy matching" or legacy fallback code that allowed it to silently ignore or map these orphaned strings—as long as the session was initialized in its standard "heavyweight" mode. **Flyway 12 (released Jan 2026)** finally purged these legacy fallbacks, making the `type` column strictly validated against known types.

### 2. The Link: Why the Transactional Lock Matters
The PostgreSQL `transactional.lock` setting is not just a toggle; it fundamentally changes Flyway's **Session Initialization Logic**:

*   **Mode A: Full Initialization (Lock = `true`)**:
    Flyway assumes it is in a "safe," standard environment. It performs a comprehensive setup, taking a transactional advisory lock. In this mode, Flyway's engine is more resilient; even in version 12, it has enough context to treat an unknown type as a "repairable" mismatch during its validation scan.
*   **Mode B: Minimalist Initialization (Lock = `false`)**:
    This mode was designed for high-concurrency environments and non-transactional DDL (like `CREATE INDEX CONCURRENTLY`). To ensure the session is as "clean" as possible and to avoid any side effects that might block concurrent operations, Flyway 12 uses a **minimalist parser** for the history table. This parser does not have access to legacy mappers or repair-logic fallbacks. If it encounters a string it doesn't recognize (like the now-deleted `SPRING_JDBC`), it immediately concludes the table is corrupted and crashes with a `FATAL` error before `repair()` can even begin.

### 3. The "Trigger" Commits: How 2.15.4 Became the Breaking Point
The reason 2.15.3 worked and 2.15.4 failed—despite both using Flyway 12—comes down to **when** the lock was disabled:

*   **In 2.15.3 (and earlier)**: The lock was disabled *programmatically* inside the migration strategy, often **after** the initial table validation had already occurred or in a way that allowed the "Full" initialization to happen first.
*   **In 2.15.4 (Commit `9145c16`)**: To fix a hang, the setting was moved to a `FlywayConfigurationCustomizer`. This Spring Boot bean configures the `Flyway` object **globally and early**.
*   **The Result**: For the first time, Flyway was forced to perform its very first read of the database history table in **Mode B (Minimalist)**. This caused the "Unknown type" error to jump from a repairable warning to a fatal crash.

### 4. Summary of Contributing Factors
*   **Legacy Debt**: Production databases containing the `'SPRING_JDBC'` type from pre-2020 installations.
*   **Library Maturity**: **Flyway 12.0.1+** strictly validating the `type` column and removing internal legacy fallbacks.
*   **Configuration Refactor (Commit `b0587fd`)**: Disabling the PostgreSQL transactional lock **early via Customizer**, which stripped away the remaining "fuzzy matching" capability of the Flyway engine.

### 5. Recommendations
*   **Immediate Fix**: The current implementation now includes an **Early Table Update** that manually executes `UPDATE flyway_schema_history SET type = 'JDBC' WHERE type = 'SPRING_JDBC';` before Flyway is even initialized. This effectively "cleans" the debt before Flyway's strict parser can see it.
*   **Citations**:
    *   [Flyway V8 Breaking Changes (Official Removal)](https://github.com/flyway/flyway/issues/3247)
    *   [PostgreSQL Transactional Lock Documentation](https://flywaydb.org/documentation/configuration/parameters/postgresqlTransactionalLock)
    *   [Community Discussion on Type Mismatches](https://stackoverflow.com/questions/53722392/flyway-how-to-replace-deprecated-springjdbcmigration-mismatch)
    *   [CredHub Issue #427 (User Report)](https://github.com/pivotal/credhub-release/issues/427)

hsinn0:
Have we reproduced the issue and verified the fix locally or in dev environment?

peterhaochen47:
> Have we reproduced the issue and verified the fix locally or in dev environment?

@hsinn0 So I tried locally but having trouble building an old credhub (in order to run the old migrations to have that legacy migration types) due to failures to download some of the old dependencies, because for that to happen, we need a pre-2020 credhub. 

At this point, I feel like the risk is low because I have, and the CI will, at least will verify that this doesn't break any existing things. And the new migration added in the PR is the exact one used in https://github.com/pivotal/credhub-release/issues/427 to resolve the issue, so that's a partial verification.

hsinn0:
OK, the code change does look fine. Let's go with this.

hsinn0 (approved):

PR #1102, opened 2026-04-23
Pin tomcat version to 10.1.54 to fix CVEs

peterhaochen47:
- to override version in our current spring-boot version (v3.5.13); can remove this pinning once spring-boot is bumped to a version containing tomcat 10.1.54 or above
(see current sprint boot tomcat version:
https://github.com/spring-projects/spring-boot/blob/v3.5.13/gradle.properties#L24)

ai-assisted=no


hsinn0:
Did you want to include this in the commit message?
```
ai-assisted=no
TASCVE...
```

I think that we should not put the internal Jira ID here at least.

peterhaochen47:
> Did you want to include this in the commit message?
> 
> ```
> ai-assisted=no
> TASCVE...
> ```
> 
> I think that we should not put the internal Jira ID here at least.

@hsinn0 fixed

hsinn0 (approved):

PR #1111, opened 2026-04-29
fix: `jdk.tls.namedGroups` jvm property value in Docker start_server.sh

### LINKED ISSUE #429: TLS connection erros with credhub-release v2.15.4
Hi CredHub Team,

we recently updated to v2.15.4 and this version caused TLS errors when trying to connect with the credhub CLI v2.9.51. Steps to reproduce:

```
$ docker run --rm -p 9000:9000 pcfseceng/credhub:2.15.4

$ export CREDHUB_DEBUG=true
$ credhub find
[DEBUG] GET /api/v1/data?path= HTTP/1.1
Host: localhost:9000
User-Agent: Go-http-client/1.1
Content-Type: application/json
Accept-Encoding: gzip

[DEBUG] An error occurred during the data request.
: Get "https://localhost:9000/api/v1/data?path=": remote error: tls: handshake failureGet "https://localhost:9000/api/v1/data?path=": remote error: tls: handshake failure
```

With v2.15.3, there is no TLS handshake error. This issue is possibly related to a Tomcat update and its TLS configuration.

Can you please check?

hsinn0:
- The JVM flag was passed with literal quote characters inside the system property value, so Tomcat 10.1.53 logged an unknown TLS group and ECDHE handshakes failed against the CredHub CLI. Use `-Djdk.tls.namedGroups=secp384r1` so the value is only the curve name.
- `docker/start_server.sh` builds argv for `java` as one double-quoted string per argument. Writing `-Djdk.tls.namedGroups=\"secp384r1\"` there put `\"` into the shell string so the JVM saw `"` characters as part of the property value, not as shell quoting. `scripts/start_server.sh` instead passes `-D...` tokens to `gradlew` using normal shell word rules, where `-Djdk.tls.namedGroups="secp384r1"` parses to a single `-D` flag whose value is `secp384r1` without embedded quotes—so the regression was specific to the Docker `java` invocation style.
- In Tomcat 10.1.53 the Connector began using jdk.tls.namedGroups as the default configured group list. See https://tomcat.apache.org/tomcat-10.1-doc/changelog.html (section "Tomcat 10.1.53 (schultz)", bullet starting with "Respect the value for the jdk.tls.namedGroups system property").
- On earlier Tomcat (e.g. 10.1.52) that linkage did not exist, so the same malformed JVM property did not affect that behavior and TLS handshakes still succeeded.
- This fixes https://github.com/pivotal/credhub-release/issues/429.

### DIFF
--- a/docker/start_server.sh
+++ b/docker/start_server.sh
@@ -8,7 +8,7 @@ declare -r application_configs
   "-Djava.security.egd=file:/dev/urandom" \
   "-Djdk.tls.ephemeralDHKeySize=4096" \
   "-Dspring.config.additional-location=${application_configs}" \
-  "-Djdk.tls.namedGroups=\"secp384r1\"" \
+  "-Djdk.tls.namedGroups=secp384r1" \
   "-Djavax.net.ssl.trustStore=/app/stores/trust_store.jks" \
   "-Djavax.net.ssl.trustStorePassword=${TRUST_STORE_PASSWORD}" \
   "-ea" \


dlinsley (approved):
Thanks for all the context and details!

PR #1112, opened 2026-05-01
Replace deprecated GenericGenerator UUID id with Hibernate IdGenerato…

hsinn0:
…rType

- Migrate EncryptionKeyCanary away from @GeneratedValue/@GenericGenerator and the custom UuidGenerator extending org.hibernate.id.UUIDGenerator.
- Introduce @AssigningRandomUuid wired to AssigningRandomUuidGenerator, a BeforeExecutionGenerator that keeps caller-assigned UUIDs and assigns UUID.randomUUID() on insert when unset. This follows Hibernate's @IdGeneratorType migration and clears deprecations flagged for removal (GenericGenerator) and the legacy UUIDGenerator subclass path.
- ai-assisted=yes, used Cursor/Auto.


hsinn0:
We found an issue with the custom uuid2 generator being used in other classes and those need to be also changed. Setting to DRAFT mode till that is addressed.

hsinn0:
> We found an issue with the custom uuid2 generator being used in other classes and those need to be also changed. Setting to DRAFT mode till that is addressed.

Correction. Actually the custom uuid2 generator was used in EncryptionKeyCanary only. The others were actually using Hibernate's built-in uuid2 generator.  I knew about this but forgot about it.  I know, it is confusing because of the same "uuid2" names is used.
See below for the (AI-generated) details:
> Before this commit, only EncryptionKeyCanary had an explicit @GenericGenerator(name = "uuid2", strategy = "org.cloudfoundry.credhub.entities.UuidGenerator") binding. The other four entities — Credential, CredentialVersionData, PermissionData, and EncryptedValue — have always carried @GeneratedValue(generator = "uuid2") without any local @GenericGenerator, so they have always resolved to Hibernate's built-in UUID strategy, not the custom UuidGenerator. This commit does not change that; it only migrates EncryptionKeyCanary away from the deleted UuidGenerator to @AssigningRandomUuid.
 
EncryptionKeyCanary is the one entity that genuinely required the "assign if already set" behavior, because its UUID is pre-assigned to match the encryption key UUID before persist(). For EncryptedValue, the key rotation path in EncryptedValueDataService.rotate() sets the UUID on a new transient object and then calls saveAndFlush() — because the UUID is non-null, Spring Data JPA routes this through em.merge() (UPDATE), not em.persist() (INSERT), so the generator is never invoked and allowAssignedIdentifiers is irrelevant.

So the change in this PR is proper as it is.

hsinn0:
BTW, the hibernate API used in the 4 other classes is deprecated, which will be handled in a different story and PR.

hsinn0:
Test steps (aI-assisted) performed:

# Upgrade Test: UUID Generator Migration

These instructions verify that the `@AssigningRandomUuid` migration (PR #1112) does not
break existing data or the key rotation path when upgrading from the previous CredHub version.

Run this test against **both PostgreSQL and MySQL** by pointing each server at the
appropriate database. The CLI commands themselves are identical for both.

## What is being tested

| Scenario | Why it matters |
|---|---|
| Existing credentials are readable after upgrade | UUID columns are unchanged; no FK breakage |
| Existing credential UUIDs are byte-for-byte identical | Generator does not re-assign on first write-back |
| Key rotation preserves `encrypted_value` UUID | `EncryptedValueDataService.rotate()` sets the UUID before save; the new generator must honour it |
| Permission UUID is stable across upgrade | `PermissionData` id unchanged; lookup by UUID still works |
| New INSERT after upgrade produces a UUID | `AssigningRandomUuidGenerator` assigns `UUID.randomUUID()` when no id is set |

---

## Prerequisites

1. `credhub` CLI is installed and on `$PATH`.
2. The database server (PostgreSQL or MySQL) is running and the schema is clean (or
   reused from a previous run — Flyway will leave it alone).
3. You have the old CredHub server binary ready to start.
4. You have the new CredHub server binary (this branch) ready to start.
5. Keep a notepad open — you will record several UUIDs during Phase 1.

### Login helper

Both phases use the same login command. The dev server at `https://localhost:9000` uses
a public UAA with fixed credentials:

```bash
credhub login \
  -s https://localhost:9000 \
  --client-name=credhub_client \
  --client-secret=secret \
  --skip-tls-validation
```

If you prefer mTLS, substitute:

```bash
credhub login \
  -s https://localhost:9000 \
  --client-cert applications/credhub-api/src/test/resources/client_cert.pem \
  --client-key  applications/credhub-api/src/test/resources/client_key.pem \
  --skip-tls-validation
```

---

## Phase 0 — Configure the old server for key rotation

Before starting the old server, ensure `applications/credhub-api/src/main/resources/application-dev.yml`
has **only one active key** (the default dev config already does this):

```yaml
encryption:
  key_creation_enabled: true
  providers:
  - provider_name: internal
    provider_type: internal
    keys:
    - encryption_password: somehardcodedencryptionpassword
      provider_type: internal
      active: true
```

Start the old server with the postgres or mysql profile, e.g.:

```bash
./scripts/start_server.sh -Dspring.profiles.active=dev,dev-postgres
# or
./scripts/start_server.sh -Dspring.profiles.active=dev,dev-mysql
```

Wait until you see `Started CredHubApp` in the log, then proceed.

---

## Phase 1 — Seed data on the old server

Login:

```bash
credhub login -s https://localhost:9000 --client-name=credhub_client --client-secret=secret --skip-tls-validation
```

### Step 1.1 — Create a password credential

```bash
credhub set -t password -n /upgrade-test/password -w "hunter2"
```

Capture its UUID:

```bash
credhub curl -p "/api/v1/data?name=/upgrade-test/password" | jq -r '.data[0].id'
```

Record: **`PASSWORD_UUID=<value>`**

### Step 1.2 — Create a value credential

```bash
credhub set -t value -n /upgrade-test/value -v "hello-world"
```

```bash
credhub curl -p "/api/v1/data?name=/upgrade-test/value" | jq -r '.data[0].id'
```

Record: **`VALUE_UUID=<value>`**

### Step 1.3 — Create a JSON credential

```bash
credhub set -t json -n /upgrade-test/json -v '{"key":"upgrade-test-value"}'
```

```bash
credhub curl -p "/api/v1/data?name=/upgrade-test/json" | jq -r '.data[0].id'
```

Record: **`JSON_UUID=<value>`**

### Step 1.4 — Create a self-signed certificate

```bash
credhub generate -t certificate \
  -n /upgrade-test/cert \
  --common-name upgrade-test \
  --is-ca \
  --self-sign
```

```bash
credhub curl -p "/api/v1/data?name=/upgrade-test/cert" | jq -r '.data[0].id'
```

Record: **`CERT_UUID=<value>`**

### Step 1.5 — Create a permission and record its UUID

The v2 permissions API requires ACLs enabled (already on in `application-dev.yml`).

```bash
credhub curl -X POST -p "/api/v2/permissions" \
  -d '{"path":"/upgrade-test/*","actor":"uaa-user:upgrade-test-actor","operations":["read","write"]}'
```

The response body contains `"uuid"`. Record it:

```bash
# re-fetch to be sure
credhub curl -p "/api/v2/permissions?path=/upgrade-test/*&actor=uaa-user:upgrade-test-actor" \
  | jq -r '.uuid'
```

Record: **`PERMISSION_UUID=<value>`**

### Step 1.6 — Confirm the old server log shows the active key UUID

In the old server log, locate a line like:

```
org.cloudfoundry.credhub.services.EncryptionKeyCanaryMapper - Creating a new active key canary
```

This confirms the server created a canary for the active encryption key. You
do not need to record it — the rotation test verifies behaviour, not this value directly.

---

## Phase 2 — Upgrade the server

### Step 2.1 — Stop the old server

`Ctrl-C` the old server process.

### Step 2.2 — Promote the second encryption key to active

Edit `applications/credhub-api/src/main/resources/application-dev.yml` to flip which key
is active:

```yaml
encryption:
  key_creation_enabled: true
  providers:
  - provider_name: internal
    provider_type: internal
    keys:
    - encryption_password: somehardcodedencryptionpassword
      provider_type: internal
      active: false          # <-- was true
    - encryption_password: correcthorsebatterystaple
      provider_type: internal
      active: true           # <-- add this
```

### Step 2.3 — Start the new server

```bash
./scripts/start_server.sh -Dspring.profiles.active=dev,dev-postgres
# or dev-mysql
```

### Step 2.4 — Wait for key rotation to complete

Watch the log. You must see this line before running Phase 3:

```
Finished encryption key rotation in … milliseconds. Details: Successfully rotated N item(s)
```

`N` should be at least 4 (one `encrypted_value` row per credential version created in Phase 1).

If you see `Found no records in need of encryption key rotation` instead, the active key
was not actually changed — go back and check Step 2.2.

---

## Phase 3 — Verify on the new server

Login:

```bash
credhub login -s https://localhost:9000 --client-name=credhub_client --client-secret=secret --skip-tls-validation
```

### Step 3.1 — Read existing credentials and check values

```bash
credhub get -n /upgrade-test/password
# Expected value: hunter2

credhub get -n /upgrade-test/value
# Expected value: hello-world

credhub get -n /upgrade-test/json
# Expected value: {"key":"upgrade-test-value"}

credhub get -n /upgrade-test/cert
# Expected: certificate block is present and parseable
```

Failure looks like: `An error occurred while retrieving the credential` or a different value.

### Step 3.2 — Verify UUIDs are unchanged

For each credential, run the API call and compare to the UUID you recorded in Phase 1.

```bash
credhub curl -p "/api/v1/data?name=/upgrade-test/password" | jq -r '.data[0].id'
# Expected: <PASSWORD_UUID>

credhub curl -p "/api/v1/data?name=/upgrade-test/value" | jq -r '.data[0].id'
# Expected: <VALUE_UUID>

credhub curl -p "/api/v1/data?name=/upgrade-test/json" | jq -r '.data[0].id'
# Expected: <JSON_UUID>

credhub curl -p "/api/v1/data?name=/upgrade-test/cert" | jq -r '.data[0].id'
# Expected: <CERT_UUID>
```

A different UUID means Hibernate re-assigned the id on rotate — the bug is present.

### Step 3.3 — Verify the permission UUID is stable

```bash
credhub curl -p "/api/v2/permissions/<PERMISSION_UUID>"
```

Expected: HTTP 200 with the permission body.

Failure looks like: `{"error":"The request includes an ACL that does not exist."}` or HTTP 404.

### Step 3.4 — Verify new INSERT assigns a UUID

```bash
credhub set -t value -n /upgrade-test/new-after-upgrade -v "post-upgrade"
credhub curl -p "/api/v1/data?name=/upgrade-test/new-after-upgrade" | jq -r '.data[0].id'
```

Expected: a non-empty UUID string (e.g. `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`).

Failure looks like: `null` or an empty string — the new generator did not assign an id.

---

## Cleanup

```bash
for name in /upgrade-test/password /upgrade-test/value /upgrade-test/json \
            /upgrade-test/cert /upgrade-test/new-after-upgrade; do
  credhub delete -n "$name"
done

credhub curl -X DELETE -p "/api/v2/permissions/<PERMISSION_UUID>"
```

Revert `application-dev.yml` to the original single-key config before the next test run.

---

peterhaochen47 (approved):

PR #1121, opened 2026-05-08
fix:  No snippets found for API doc generation

hsinn0:
- Changed to use TestInfo to provide accurate test method names for REST documentation.
- Updated @BeforeEach setUp methods across 14 controller test classes to inject TestInfo.
- Changed restDocumentation.beforeTest() to use testInfo.testMethod.get().name instead of javaClass.simpleName.
- Ensureed REST documentation snippets are generated with unique identifiers per test method rather than the class name.

ai-assisted=yes
TNZ-100436


peterhaochen47 (dismissed):

peterhaochen47 (approved):

hsinn0:
Forgot to check ktlint in the fist commit.

PR #1140, opened 2026-06-11
skip major and minor bumps for springboot

dlinsley:
spring boot published `4.1.0` and `4.0.7` recently

`4.1.0` has some breaking change.  would like dependabot to bump us to `4.0.7` now instead of `4.1.0`.


hsinn0 (commented):
OK, if we are going to do this, can we create a story to bump to 4.1 (and undo this change) soon?

dlinsley:
Created story as TNZ-109817

hsinn0 (approved):

PR #1154, opened 2026-06-25
fix: percent-encoded path bypass of ManagementInterceptor loopback gate

hsinn0:
- A request to /%6Danagement (or any other percent-encoded equivalent) has requestURI = "/%6Danagement", which is not string-equal to "/management", so the loopback check was silently skipped
- Spring Security's PathPatternRequestMatcher and Tomcat's servlet mapping both operate on the decoded path, so the request was still permitted and dispatched to ManagementController.updateManagementRegistry(), allowing an unauthenticated remote caller to toggle CredHub's read-only mode
- Fix: decode requestURI with UriUtils.decode() (stripping context path first) before any comparison, mirroring what Tomcat's CoyoteAdapter.UDecoder does when producing servletPath; this works in both production (Tomcat decodes) and MockMvc tests (requestURI is always set, servletPath is not)
- Apply the same decoded path variable to the read-only mode exemption checks for /management and /interpolate so encoded variants of those paths are also correctly handled

ai-assisted=yes
TNZGOV-4101


dlinsley (approved):
Thanks for adding a respective test!

PR #1157, opened 2026-06-29
fix: hanging test

hsinn0:
- Replace Object lock/wait/notify synchronization with CountDownLatch.

ai-assisted=yes
TNZ-115123

### DIFF
--- a/components/encryption/src/test/java/org/cloudfoundry/credhub/services/RetryingEncryptionServiceTest.java
+++ b/components/encryption/src/test/java/org/cloudfoundry/credhub/services/RetryingEncryptionServiceTest.java
@@ -2,6 +2,8 @@ package org.cloudfoundry.credhub.services;
 
 import java.security.ProviderException;
 import java.util.UUID;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import javax.crypto.IllegalBlockSizeException;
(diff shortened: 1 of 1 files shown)


components/encryption/src/test/java/org/cloudfoundry/credhub/services/RetryingEncryptionServiceTest.java
```diff
@@ -462,28 +464,24 @@ private class RacingRetryingEncryptionServiceForTest extends RetryingEncryptionS
...
 
     @Override
     public void setNeedsReconnectFlag() {
       try {
         if (Thread.currentThread().equals(firstThread)) {
           secondThread.start();
-          synchronized (lock) {
-            lock.wait(); // pause the first thread
-          }
+          latch.await(10, TimeUnit.SECONDS); // pause the first thread until secondThread signals
```
dlinsley: question: Would we want to capture the result here to determine if the latch counted down (`true`) or the timeout occured (`false`)?  I think we would want to fail the test of the timeout occured?

Also, is the `Thread.sleep(10)` still needed with the `await`?

hsinn0: Updated the test code to check for the latch signal.

> is the Thread.sleep(10) still needed with the await?

Technically no, but I just leave the 10 ms sleep as is just in case.

dlinsley (approved):

PR #966, opened 2025-08-14
Adding duration paramater for regenerate

benjaminguttmann-avtq:
This change introduces the possibility to provide a duration parameter when regenerating a certificate.

As this is my first PR to a Java project I am open for suggestions :D.

Thank you.

PR #1161, opened 2026-07-10
Allow to adjust cert duration on bulk-regenerate

benjaminguttmann-avtq:
This PR should introduce the possibility to adjust the duration of certificates using bulk-regenerate.

Follow up of https://github.com/cloudfoundry/credhub/pull/966

### DIFF
--- a/backends/credhub/src/main/kotlin/org/cloudfoundry/credhub/regenerate/DefaultRegenerateHandler.kt
+++ b/backends/credhub/src/main/kotlin/org/cloudfoundry/credhub/regenerate/DefaultRegenerateHandler.kt
@@ -71,7 +71,10 @@ class DefaultRegenerateHandler(
         return CredentialView.fromEntity(credentialVersion, concatenateCas, true)
     }
 
-    override fun handleBulkRegenerate(signerName: String): BulkRegenerateResults {
+    override fun handleBulkRegenerate(
+        signerName: String,
+        duration: Int?,
+    ): BulkRegenerateResults {
         auditRecord.requestDetails = BulkRegenerateCredential(signerName)
 
         verifyRegeneratePermissions(signerName)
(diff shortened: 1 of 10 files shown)


peterhaochen47:
Thanks for the contribution. This change is not wrong per se, but I have some reservations about how this complicates our relatively simple design. 

1. The use case is solvable today, with better precision

The scenario this PR addresses — regenerating certificates with updated durations — can already be accomplished by using /v1/certificates/:id/regenerate with a duration override on each cert individually (thank you for #966). This is a better, more precise control, because you can decide to set different new durations for leafs vs. cas on a per cert basis. While this requires more calls upfront, duration changes are inherently a rare, intentional act: once you've corrected the duration on a cert, future regenerations (including bulk) simply reuse the stored parameter. The one-time setup cost is a reasonable trade-off for the precision it affords.

2. The CA vs. leaf distinction creates some confusion

The group of certs being bulk-regenerated could contain both CAs and leafs (since a CA can sign another CA, aka the intermediate CA). CredHub already models CA and leaf certificates as having distinct duration semantics at the server level (see the server level configs: `certificates.ca_minimum_duration_in_days` vs. `certificates.leaf_minimum_duration_in_days`). A single duration parameter applied uniformly across a CA tree collapses that distinction. Furthermore, the single-cert regenerate endpoint today returns `duration_overridden` and `duration_used` in its response so callers know exactly what was applied. Bulk-regenerate's response is a flat set of names — there is no per-cert audit trail. With this PR, a caller can influence duration but cannot observe the outcome. That's a meaningful observability regression relative to what we provide on the single-cert path. If we try to fill this gap, it'll further complicate this originally simple endpoint or break the request response contract.

3. This is sort of not the original intent of the bulk regenerate endpoint

The documented intent of the regenerate endpoints is to "generate new values for credentials using the same parameters as the stored value." This is precisely what makes bulk-regenerate safe to use: its blast radius is large, but it is bounded — it doesn't alter any parameters, only reissues. Introducing duration complicate this simple intent & enlarge the blast radius. Once the endpoint accepts one generation parameter, it becomes harder to reason about what "bulk regenerate" means, and it sets a precedent for other parameter overrides (key_length, key_usage) that we may not want to follow.

benjaminguttmann-avtq:
Understood. Thank you for the explanation :)

PR #1175, opened 2026-08-03
Update gradle-versions-plugin specifier

hsinn0:
- Move gradle-versions-plugin to settings.gradle using io.github.ben-manes.versions.settings.
- Per https://github.com/ben-manes/gradle-versions-plugin/releases/tag/v0.56.0.

ai-assisted=yes
TNZ-124015

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,10 +37,6 @@ buildscript {
     }
 }
 
-plugins {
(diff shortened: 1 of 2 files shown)


dlinsley (approved):
Thank You!

PR #1182, opened 2026-08-14
pin httpcore5 to 5.4.3

dlinsley:
this is beyond the existing spring-boot constraint to resolve CVE-2026-54399


hsinn0 (changes requested):
Can we have a comment that explains the reason for the version pinning?

hsinn0 (approved):

PR #1187, opened 2026-08-26
remove: dependency version pinning of httpcore5.version to 5.4.3

hsinn0:
- Manually revert https://github.com/cloudfoundry/credhub/pull/1182/.
- As we have bumped spring-boot to 4.1.1 and its managed dependency version of httpcore5.version is 5.4.3.


dlinsley (approved):
Thanks!

PR #1198, opened 2026-09-16
Luna 10.9 junit

hsinn0:
Optional JUnit test connecting to a HSM partition and accompanying document


docs/running-luna-compatibility-harness.md
```diff
@@ -0,0 +1,291 @@
...
+> :warning: **DANGER: RISK OF PERMANENT DATA LOSS**
+>
+> Luna HSM partitions enforce a **bad-login attempt counter** (typically between 3 and 10 attempts). If consecutive failed login attempts exceed the partition threshold, the HSM **permanently locks or zeroizes (erases) the partition**, destroying all keys and data!
+
+### Mandatory Pre-Flight Verification
+* **NEVER** run the Gradle test tasks with unverified credentials or typos in `LUNA_HSM_PARTITION_PASSWORD`.
+* **ALWAYS** verify connectivity and credentials out-of-band using `lunacm` before invoking Gradle:
+  ```bash
+  lunacm
+  lunacm:> partition login -partition <PARTITION_NAME> -password <PARTITION_PASSWORD>
```
peterhaochen47: This is failing for me:
```
lunacm:>partition login -partition <partition_label> -password <partition_password>
Subcommand "login" not found.
```

peterhaochen47: This instead is working:
```
lunacm:> slot list
lunacm:> slot set -slot <slot_id>
lunacm:> role list
lunacm:> role login -name co -password <co password>
lunacm:> partition showinfo
lunacm:> role logout
```

peterhaochen47: > In modern Luna (7.x and 10.x), the subcommand partition login still exists in LunaCM, but only when the active slot is an HA virtual slot (e.g., when multiple HSM appliances are clustered into a virtual HA partition)

So it seems like the `partition login` [command](https://thalesdocs.com/gphsm/luna/7/docs/network/Content/lunacm/commands/partition/partition_login.htm) could work under some condition, when HA is configured.

hsinn0: Right, it is `role login`. `partition login` was not available subcommand there. I will fix the doc.

hsinn0: BTW, can you edit out the real password from one of your comments, @peterhaochen47?

hsinn0: Never mind. I was able to edit your comment. I did not know I could do that.

docs/running-luna-compatibility-harness.md
```diff
@@ -0,0 +1,294 @@
...
+     CardProc = {
+       ServerCAFile = /usr/safenet/lunaclient/cert/server/server.pem;
+     }
+     ```
+3. **Mutual TLS Registration**:
+   * HSM server CA registered with `vtl addServer`.
+   * Client certificate created and registered on the Luna appliance via `client register`.
+   * Client assigned to the partition via `client assignPartition`.
+4. **Network Connectivity**:
+   * TCP port **1792** (NTLS) must be open and reachable between the host running the tests and the Luna HSM appliance.
```
peterhaochen47: This NTLS (port 1792) requirement is only for physical on-prem HSM device right (which we have not actually run against yet)? Not the cloud HSM (which this harness has actually been validated against). Though theoretically this harness can be run against both; we just haven't tried the former yet. It might be worth a note explaining this NTLS requirement is only for on-prem HSM?

> Suggested Improvement
Add a short note clarifying the two network topologies:
On-Prem Physical Appliance: Uses NTLS over TCP port 1792 (vtl verify, ServerCAFile).
Thales DPoD (Cloud HSM): Uses REST/HTTPS over TCP port 443 via plugins/libcloud.plugin with OAuth client credentials.

docs/running-luna-compatibility-harness.md
```diff
@@ -0,0 +1,294 @@
...
+
+This document describes how to configure, execute, and interpret CredHub's opt-in Luna Hardware Security Module (HSM) compatibility test suite.
+
+---
+
+## 1. Overview and Purpose
+
+CredHub supports Thales/SafeNet Luna HSM as an external cryptographic provider via Java Cryptography Architecture (JCA) integration. The test suite defined in `LunaHsmCompatibilityTest.java` validates CredHub's Luna JCA integration directly against a physical HSM partition at the code level.
+
+The compatibility harness verifies:
```
peterhaochen47, line 11: small non-blocking suggestion: might be worth a note listing the parts of our eventual feature that this harness does not cover, so the reader has a better sense of the scope of this harness & not overestimate the confidence that it establishes.

docs/running-luna-compatibility-harness.md
hsinn0: Restructured the section to show options for using physical hsm service vs. cloud hsm service.

docs/running-luna-compatibility-harness.md
hsinn0: As this doc is about how to run the JUnit tests, I think that I will just keep it as it is. Maybe, after we have complete implementation later, we can come back to this doc to adjust further accordingly.

peterhaochen47 (approved):

